PR #15, opened 2018-09-11
Cnv template mechanism

fabiand:


templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+  labels:
+    # The UI can show all possible template.cnv.io/* values in a nice way
+    # and let the user filter down the available templates to the one
+    # the user actually wants:
+    # A single selected template only means no conflicts and no smart
+    # merging code. This has to be done using labels to allow efficient
+    # searching.
+    # The format has the following meaning:
+    # {os,flavor,size}.template.cnv.io/{value}: {version} or just true
+    os.template.cnv.io/windows2k12r2: 201809
```
fabiand: I think we don't need a version here, we should just version the whole template.

And this is the case as soon as it's stored in K8s.

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+    # A single selected template only means no conflicts and no smart
+    # merging code. This has to be done using labels to allow efficient
+    # searching.
+    # The format has the following meaning:
+    # {os,flavor,size}.template.cnv.io/{value}: {version} or just true
+    os.template.cnv.io/windows2k12r2: 201809 
+    os.template.cnv.io/windows8: 201809 
+    os.template.cnv.io/windows7: 201809 
+    flavor.template.cnv.io/minimal: 201809
+    # io.cnv.vm.size not specified means all?
```
fabiand: We should require all the main parts to be set, os, size, workload.

If we want to say "ALL", then we might want to use such a special keyword i.e. `size: ALL`

The names "windows8" should rather be the libosinfo identifier whhich provides unique ids for all major osses. I also have the mipression that the os should be a value, rather than a key.

All in all I'm not yet sure about how the fields shoudl excatly look.

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+      STORAGE_CLASS:
+      NETWORK_CLASS:
+
+  labels:
+    # This labels will link the VM to a template that was
+    # used to create it. This can be then used by the user
+    # or UI for recomputing the VM.spec using updated
+    # template. A VM without this label can be considered
+    # "baked" and not linked to any template.
+    vm.cnv.io/template: windows
```
fabiand: IIRC OCP Templatea already add ownership informations. This should be used if possible.

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+apiVersion: v1
+kind: VirtualMachine
+metadata:
+  name: windows-10-1
+  annotations:
+    # Data used for the UI, ignored by kubevirt
+    # Arbitrary format as needed to be able to
+    # repopulate the UI or the template processor
+    # and get the same output
+    template.cnv.io/original:
```
fabiand: I _think_ I have seen that OCP Templates already store the original parameter values. Somethig to be checked.

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+    # Another extension useful for CNV aware tooling that allows
+    # expressing additional validation rules for this template
+    template.cnv.io/validate:
+    - path: "/VM/spec/domain/guest/memory"
+      min: 3
+      max: 4
+
+    # Another extension for hinting at which elements should be
+    # considered editable
+    template.cnv.io/editable:
```
fabiand: A key thing of this is that I'd expect the UI to modify the VM.spec part directly after the VM got created.
The UI would use the informations above for knowing _what_ to edit, and to know _what_ valid input is.

/cc @jelkosz

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+  from: '[A-Z0-9]{8}'
+  generate: expression
+  name: MEMORY_SIZE
+  value: 1Gi
+
+objects:
+# The full VM template with placeholders for either scalars like memory
+# or yaml structures like disks
+- apiVersion: v1
+  kind: VirtualMachine
```
fabiand: In this exampel teh VM is pretty small.

We either provide all the details using presets or directly within the template.

The mental model if we have, template, vm, vmi only is simpler than with presets.

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+  annotations:
+    description: "Basic windows 10 template"
+    
+    # This set of annotations is meant as a hint only, it is
+    # not going to be processed by the stock openshift templating
+    # mechanism. CNV UI can and should use it though.
+    # The information encoded here informs the tooling (UI)
+    # about names of devices that are to be used as templates
+    # for adding additional disk, networks, volumes and such
+    defaults.template.cnv.io/disk: default-disk
```
fabiand: The question is if these …

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+    defaults.template.cnv.io/volume: default-volume
+    defaults.template.cnv.io/network: default-network
+
+    # Replacements useful for finishing the transformation
+    # This is an extension we could use for specifying common
+    # fields for all disks/nics. Requires extra implementation
+    # that will do the find and replace using a syntax documented
+    # here: https://kubernetes.io/docs/reference/kubectl/jsonpath/
+    template.cnv.io/replace:
```
fabiand: … and tehse and other metadata informations should be part of the Template or fo the VM.

A benefit of having it on the VM is that the UI only needs to look at VMs for editing it.
And does not need to reconcile with the original Temaplte.

Also: A template can contain more than one entity, this needs to be reflected in the json paths .. ias it could be possibe that we also include a service, whichmight need validation ifnromations as well.

templates/README.md
```diff
@@ -0,0 +1,221 @@
...
+    # template. A VM without this label can be considered
+    # "baked" and not linked to any template.
+    vm.cnv.io/template: windows
+
+# The requested state of the VM that will always match what
+# the user asked for exactly. When the UI pushes edits it
+# simply replaces the whole spec: with the current version
+# the user saved
+spec:
+  domain:
```
fabiand: A UI would permit editing this tree here, based on the informations in ` template.cnv.io/editable:`

templates/README.md
MarSik: We want all VMs to be based on a template if created this way and we should provide a Blank template as well (possibly with just the OS bits). And I agree that the mental model of template -> vm -> vmi is simpler.

templates/README.md
fabiand: These are annotations and here we've got the complete freedom. Take a look at these examples: https://docs.okd.io/latest/dev_guide/templates.html#writing-description

templates/README.md
MarSik: This is just a hint for the UI to know how a disk / nic /.. should look like. We can do without it if we can provide the necessary details like default disk type (scsi, virtio) using some other VM.spec attribute.

templates/README.md
MarSik: Yes, but not in the case of updated template.

templates/README.md
MarSik: I will check that, but yes, we want to reuse as much as possible.

templates/LABELS.md
```diff
@@ -0,0 +1,59 @@
...
+- os.template.cnv.io/fedora28
+- os.template.cnv.io/fedora27
+- os.template.cnv.io/fedora26
+- os.template.cnv.io/fedora25
+- os.template.cnv.io/fedora24
+- os.template.cnv.io/fedora23
+
+### CentOS
+
+- os.template.cnv.io/centos7.0
```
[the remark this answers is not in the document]
MarSik: The libosinfo upstream version does not have any newer ones.. https://gitlab.com/libosinfo/osinfo-db/tree/master/data/os/centos.org

But I will add RHELs too.

templates/LABELS.md
```diff
@@ -0,0 +1,59 @@
...
+- os.template.cnv.io/centos6.9
+- os.template.cnv.io/centos6.8
+- os.template.cnv.io/centos6.7
+
+### Microsoft Windows
+
+- os.template.cnv.io/win2k8r2
```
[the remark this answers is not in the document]
MarSik: Some people might still run XP, just saying...

templates/LABELS.md
```diff
@@ -0,0 +1,59 @@
...
+- os.template.cnv.io/win2k8
+- os.template.cnv.io/win10
+- os.template.cnv.io/win8.1
+- os.template.cnv.io/win8
+- os.template.cnv.io/win7
+- os.template.cnv.io/winvista
+- os.template.cnv.io/winxp
+
+## Workload profiles
```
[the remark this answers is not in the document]
MarSik: - we currently have no distinction between server and desktop in CNV, but I can obviously do that easily
- sap-hana is an example and we do not have to use it just yet
- high-performance is CPU intensive load vs. io-intensive is disk performance, those two do not have to be the same, right?

templates/LABELS.md
```diff
@@ -0,0 +1,59 @@
...
+- workload.template.cnv.io/sap-hana
+
+## Flavors
+
+- flavor.template.cnv.io/tiny
+- flavor.template.cnv.io/small
+- flavor.template.cnv.io/medium
+- flavor.template.cnv.io/large
+- flavor.template.cnv.io/xlarge
+- flavor.template.cnv.io/dedicated-host
```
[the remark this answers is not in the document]
MarSik: It came up in one of the design discussions, we might want to allow easy sizing based on the host it ends up at. Think SAP or the main Oracle database for a company.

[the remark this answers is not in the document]
michalskrivanek: eh, having unpredictable size doesn't sound very useful. Again, I would only add stuff we have some use for right now, and add fancier things later

templates/LABELS.md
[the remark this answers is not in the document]
michalskrivanek: I wouldn't give anyone any ideas yet. We should build workloads as we go. Too many would be confusing.

if high performance == cpu intensive and we're thinking of adding i/o instensive, doesn't it make sense to call "cpu intensive" or some other word, rather than "high performance" which is too generic and makes sense when you only have one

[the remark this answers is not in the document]
MarSik: Sure.

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    template.cnv.io/replace:
+    - path: "/VM/disks/*/type"
+      value: virtio
+
+    # Another extension useful for CNV aware tooling that allows
+    # expressing additional validation rules for this template
+    # The jsonpath root is the objects: element of the template
+    # XXX There is currently an issue with annotations not supporting
+    #     list values
+    template.cnv.io/validate:
```
jelkosz: you need to somehow express also the expected type since the min-max has different meaning for string and int

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    - path: "/VM/spec/domain/guest/memory"
+      min: 3
+      max: 4
+
+    # Another extension for hinting at which elements should be
+    # considered editable. The content is a line separated
+    # list of jsonpath selectors.
+    # The jsonpath root is the objects: element of the template
+    template.cnv.io/editable: |
```
jelkosz: I don't think this is actually needed. UI knows which parts of the spec it can edit and I don't think it makes sense to configure some parts where we want to disallow it.

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    # OS names should match the libosinfo identifiers
+    # flavors are tiny, medium, large, etc.
+    # workloads are generic, high-performance, io-intensive,
+    #               oracle-db, sap-hana...
+    os.template.cnv.io/windows2k12r2: "true"
+    os.template.cnv.io/windows8: "true"
+    os.template.cnv.io/windows7: "true"
+    workload.template.cnv.io/minimal: "true"
+    workload.template.cnv.io/io-intensive: "true"
+    # flavor.template.cnv.io/* not specified means all
```
jelkosz: I dont think we need it? I think it would be easier if the UI would simply filter whatever is behind the '/'. If we will have this notion of "all" and "all but" it means we need to have an independent database of the values which we would need to maintain etc

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    openshift.io/support-url: "https://access.redhat.com"
+    iconClass: icon-windows    
+
+    # This set of annotations is meant as a hint only, it is
+    # not going to be processed by the stock openshift templating
+    # mechanism. CNV UI can and should use it though.
+    # The information encoded here informs the tooling (UI)
+    # about names of devices that are to be used as templates
+    # for adding additional disk, networks, volumes and such
+    defaults.template.cnv.io/disk: default-disk
```
jelkosz: what does this actually mean? The bootable disk?

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    defaults.template.cnv.io/volume: default-volume
```
jelkosz: same question

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    defaults.template.cnv.io/network: default-network
```
jelkosz: same question

templates/LABELS.md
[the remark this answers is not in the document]
MarSik: You do not have to show the IDs directly. The IDs here match libosinfo database IDs so feel free to display it as needed.

templates/README.md
```diff
@@ -0,0 +1,254 @@
...
+    # list of jsonpath selectors.
+    # The jsonpath root is the objects: element of the template
+    template.cnv.io/editable: |
+      /objects[0].spec.template.spec.domain.cpu.cores
+      /objects[0].spec.template.spec.domain.resources.requests.memory
+      /objects[0].spec.template.spec.domain.devices.disks
+      /objects[0].spec.template.spec.volumes
+      /objects[0].spec.template.spec.networks
+
+  labels:
```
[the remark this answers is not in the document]
MarSik: For what purpose? To not display all templates in the system? So maybe a label like template.cnv.io/type: base would be the right one?

templates/README.md
MarSik: The validation format needs some more thinking, as the value of annotation has to be a plain string. So we need to invent the format. My proposal would be one validation rule per line, something like

json_path type=int max=3
json_path type=string max_length=5 regexp=""

templates/README.md
MarSik: Fabian indeed wanted to be able to limit the editable areas, there is a use case for that. Think about GCE or AWS where you only select the template (t2.small), but are not allowed to customize it.

templates/README.md
MarSik: Or just prepare the in memory list of allowed fields by loading all templates first.

templates/README.md
MarSik: The disk entity you should use as a template for constructing all added disks. Basically prepare a disk in UI and add all specified fields from the template (like the bus: virtio for example). This is needed to support different default drivers on different OSes.

templates/README.md
MarSik: same answer :)

templates/README.md
MarSik: and same answer again :)

templates/LABELS.md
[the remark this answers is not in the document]
fabiand: The libosinfo database is splitting them. IOW the UI can use the libosinfo identifier to retrieve a lot of more data.

Example of spltted version and family:
`osinfo-query -f id,family,version os`

templates/README.md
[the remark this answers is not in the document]
fabiand: Godo queastion: Why would you need to know who created it? if the template is valid and has the right metadata, then the UI should not bother.

templates/README.md
fabiand: We could also look at OpenAPI for speciying valid input.

templates/README.md
fabiand: @MarSik why instead not:

```
os: windows7
flavor: large
workload: io-intensive
```

I'd rather have three templates which are slightly different i.e. to reflect large, small, medium, in favor of a simple tempalte and nice labels.

templates/README.md
MarSik: Sure, but bear in mind the UI layer will have to implement this using javascript and that there will be no yaml format/indentation validation from IDEs, because the template will treat that as a string.

templates/README.md
MarSik: Because you will not have three files, but thirty. And rhel7.0 is going to be identical to rhel7.1, rhel7.2, ... , rhel7.5.

templates/README.md
MarSik: Btw, we could also express this logic using the replacements that were proposed as extensions. It will just require specifying them now and might result in less readable template. And the disk there would have to be removed by the UI, but it is used in test automation.

templates/LABELS.md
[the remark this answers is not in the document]
fidencio: Regarding the osinfo-db and CentOS, CentOS has decided to stuck with 7 and not do 7.1, 7.2, 7.3 updates. They just release a new ISO of 7 and that's it.

About RHELs, @MarSik, there are no released versions missing.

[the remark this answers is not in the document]
MarSik: I know, I already added RHELs.

templates/README.md
fabiand: Yes, yes, yes ...

templates/README.md
MarSik: The is also default-volume, just two lines below :)

MarSik: And this is supposed to define what kind of disk is supported by the OS mainly in terms of bus (ide, scsi, sata, virtio, ...)

templates/README.md
fabiand: Actually - It looks like this is done by the UI, or a broker.

But plain temapltes `oc process  | oc apply -f -` will not carry ownership informations.
To be honest, it's not ownership, but object reference would be helpful.

Eventually something to enhance in template component.

templates/README.md
fabiand: OpenAPI is field validation, not tied to a language.

MarSik: OpenAPI/Swagger is much more than that. The UI will have to implement the parser for the format (or use existing js library) so something simpler that uses json path might be a better fit, because we use it elsewhere and it can match multiple disks with single rule (wildcards).

templates/README.md
michalskrivanek: would be best if it clearly states that it's the default guest device type. so..just your last comment transformed into the readme:)

MarSik: Good idea. I wonder if we need to distinguish the default template based on disk type - HDD vs. CD

michalskrivanek: not really. That's not what you're commonly adding to VM. Sure, the driver may be different for Win vs Lin (SATA vs virtio-scsi) but you would put that straight to the template I suppose (as you'd want to create an empty CDROM all the time) and then the backend code which would add another one for e.g. cloud-init just need to use the same as the existing. I don't think that logic exists, but it's not too difficult.

Alternatively you could add a default for CDs...but...ideally we don't want to explode there either so unless it's really needed I would try to avoid introducing more

jelkosz: okey, I've got your intention now :) So how would you express the list of available values in the default disk? e.g. it supports both virtio-scsi and ide, but the default is virtio-scsi?

templates/LABELS.md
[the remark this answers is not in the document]
jelkosz: @fabiand and how would I use it from the UI? Write a service which will expose it using some API?

templates/README.md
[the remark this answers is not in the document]
jelkosz: It is not about who created it but what is the purpose of it.

The combination of OS/workload profile/flavor always has to match exactly one template (IIUC). This whole concept here is about having some common configurations. It is not a template in the oVirt sense where the user takes a whole VM and turns it into a template. So in order to differentiate between pure configurations and user templates we should have some flag differentiating.

From UI perspective (design: https://github.com/kubevirt/web-ui-design/pull/5), differentiate between this two flows (ignoring storage/network in this flows):

1: use templates as described here
 - pick a provision source like PXE
 - pick the OS/workload/flavor
 - create

2: use the ovirt style templates
 - pick a provision source: template (e.g. custom template - whole copy of a VM)
 - the os/workload profile will not be editable, the flavor can only be changed to "custom" or left as it was
 - if picked "custom" flavor, you can edit memory/cpus
 - create

What is written here all makes sense for both, just that the UI needs to know if it is a configuration template (e.g. osinfo like) or an ovirt style template.

templates/LABELS.md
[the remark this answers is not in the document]
MarSik: @jelkosz service is the answer to everything :) or you could start by pre-generating the id to OS / version map during build time

templates/README.md
MarSik: The disk specification can be quite different based on what disk, volume or nic is being used. The intention here is to just merge the specified template with what you create in the UI. If the OS supports both then the template will either prefer one or the ui is free to provide any value (or none to use kubevirt defaulting).

MarSik: To make it clear, the list of editable fields for disks is not going to be part of the template. The UI will need to have some knowledge about how disks are specified. The template is just supposed to be applied on top of what the UI prepares.

michalskrivanek: so it's fine that the disk structure is going to be written by UI, that's expected, but for validation&display UI needs to know valid values for that OS/kubevirt version

templates/README.md
fabiand: Along those lines: We could see that we get "linux families" or "tribes" which the reflect
"ubuntu", "rhel", "fedora"
which canbe a generic base.

templates/README.md
MarSik: Right, we can prepare the validation rules for those. As soon as we figure out the format..

templates/LABELS.md
```diff
@@ -0,0 +1,85 @@
...
+- os.template.cnv.io/rhel6.0
+- os.template.cnv.io/rhel6.1
+- os.template.cnv.io/rhel6.2
+- os.template.cnv.io/rhel6.3
+- os.template.cnv.io/rhel6.4
+- os.template.cnv.io/rhel6.5
+- os.template.cnv.io/rhel6.6
+- os.template.cnv.io/rhel6.7
+- os.template.cnv.io/rhel6.8
+- os.template.cnv.io/rhel6.9
```
fidencio: I'd also add rhel6.10 and centos6.10 as they're already released

templates/LABELS.md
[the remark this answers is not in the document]
fidencio: Related to this (but thinking about other OSes). Do we really want to provide any kind of template/installation info for systems that have reached their EOL?
Maybe Windows is a different case, but IMO we shouldn't be providing, for instance:
centos < 6.10
fedora < 27
...

templates/README.md
fidencio: Can we actually have some inheritance of workload/profiles?
I mean, we can set, for instance:
```
os: fedora21
flavor: larger
workload: io-intensive
```

And for the next one's we could have something like:
 ```
os: fedora22
derives-from: fedora21
```

Those could give us, for free, the flavor workload we want.
This is basically what we do on osinfo-db regarding the supported devices.

Of course, going for families is also an options.

File: templates/fedora28.yaml [hunk 22]
```diff
@@ -1,57 +1,61 @@
...
   spec:
     running: false
     template:
-      metadata:
-        labels:
-          kubevirt.io/os: fedora28
       spec:
         domain:
+          cpu:
+            cores: 2
```
fidencio: I understand this is still a POC, but why do we have two here? Shouldn't this also come with a comment explaining from where it comes from or just use the recommended values from osinfo-db?

templates/fedora28.yaml
```diff
@@ -1,57 +1,61 @@
...
+          resources:
+            requests:
+              memory: 2G
```
fidencio: Same here ...

fidencio: And in a few other places down below.

templates/rhel75-hp.yaml
```diff
@@ -0,0 +1,84 @@
...
+    template:
+      spec:
+        domain:
+          cpu:
+            cores: 2
+            dedicatedCpuPlacement: true
+            model: host-model
+          resources:
+            requests:
+              memory: 4G
```
fidencio: Here again, I do believe that in order to help us in the future we should at least have a nice comment on why we decided to go for 4GB instead of something else.

fidencio: And the comment also applies for the number of cores.

templates/README.md
```diff
@@ -52,24 +52,32 @@ metadata:
...
+    #     list values
     template.cnv.io/replace:
     - path: "/VM/disks/*/type"
       value: virtio
 
     # Another extension useful for CNV aware tooling that allows
     # expressing additional validation rules for this template
     # The jsonpath root is the objects: element of the template
+    # XXX There is currently an issue with annotations not supporting
+    #     list values
```
fidencio: This should go in the commit message. Having to dig into the code to find out which are the issues that have to be fixed in the template is not so practical.

Also, couldn't this commit be squashed to the one adding the templates?

templates/README.md
```diff
@@ -117,7 +117,7 @@ parameters:
 objects:
 # The full VM template with placeholders for either scalars like memory
 # or yaml structures like disks
-- apiVersion: v1
+- apiVersion: kubevirt.io/v1alpha2
```
fidencio: Shouldn't these be squashed to the first commit as well?

templates/README.md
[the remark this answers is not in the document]
jelkosz: @MarSik so, going back to this :) We need a label to be able to say if it is a configuration template or a VM template. Would you be okey to call it: "template.cnv.io/type: vm/config"?

We will need to agree on this so we will be able to update the https://github.com/ansibleplaybookbundle/import-vm-apb/tree/master/templates to be able to consume templates imported using import-vm-apb.

templates/ubuntu1804.yaml
```diff
@@ -34,14 +34,6 @@ metadata:
...
 objects:
 - apiVersion: kubevirt.io/v1alpha2
-  kind: VirtualMachineInstancePreset
-  metadata:
-    name: ubuntu1804
-  spec:
-    selector:
-      matchLabels:
-        kubevirt.io/os: ubuntu1804
-- apiVersion: kubevirt.io/v1alpha2alpha2
```
fidencio: Would be nice to have in the commit message a nice explanation on why this is breaking the ubuntu template.
Is this material to be squashed to one of the previous commits?

templates/README.md
MarSik: No inheritance in the shipped templates. That would mean all the runtime tooling needs to load the whole chain. Instead you can add multiple OSes to the same template. Also the limitation here is in what Openshift template mechanism supports. The templates must work in CLI and CI automation environments with what we currently have. 

However, the plan is to compile the final templates from snippets using some kind of source code level templating and compile them into the final form before release.

templates/fedora28.yaml
MarSik: Because I took and updated the existing templates. I do not like the sizing aspect of those much either. We can strip that and ship with size: custom label too.

MarSik:
I will squash the patches before merge, but it is not the preferred git workflow as it requires forcing the push or constant rebases.

templates/README.md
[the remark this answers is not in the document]
MarSik: Works for me. template.cnv.io/type is OK and use whatever type you want. I am not sure we have to complicate it though "vm/config" seems redundant, "base" or "config" is probably enough.

templates/LABELS.md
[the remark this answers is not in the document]
fabiand: @jelkosz there are many ways. There could be a service, or you compile the model into some JSON objects.

templates/README.md
```diff
@@ -0,0 +1,276 @@
...
+    # and is not going to be processed by the stock openshift templating
+    # mechanism. CNV UI can and should use it though.
+    # The information encoded there informs the tooling (UI)
+    # about names of devices that are to be used as templates
+    # for adding additional disk, networks, volumes and such
+
+    # The goal of default disk is to define what kind of disk
+    # is supported by the OS mainly in terms of bus (ide, scsi,
+    # sata, virtio, ...)
+    defaults.template.cnv.io/disk: default-disk
```
fabiand: 1. Let's call it …/disk-bus
2. Let's use a sane example (i.e. virtio)

templates/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+    # The goal of default volume is to be able to configure mostly
+    # performance parameters like caches if those are exposed
+    # by the underlying volume implementation.
+    defaults.template.cnv.io/volume: default-volume
```
fabiand: cache is part of the disk struct.

Thus I would rather see it as part of the …/disk annotation above.

Do we need it right now?

templates/README.md
fabiand: Would this field always just contain one value (the bus)?

templates/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+    # The goal of default network is similar to default-disk
+    # and should be used as a template to ensure OS compatibility
+    # and performance
+    defaults.template.cnv.io/nic: default-nic
```
fabiand: IIUIC this will then define the default nic bus? i.e. virtio?

if so, then lte's call it that way …/nic-bus

templates/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+    # The goal of default network is similar to default-volume
+    # and should be used as a template that specifies performance
+    # and connection parameters (L2 bridge for example)
+    defaults.template.cnv.io/network: default-network
```
fabiand: What parameteres are expected to be set here?

templates/README.md
```diff
@@ -0,0 +1,276 @@
...
+    # This is an extension we could use for specifying common
+    # fields for all disks/nics. Requires extra implementation
+    # that will do the find and replace using a syntax documented
+    # here: https://kubernetes.io/docs/reference/kubectl/jsonpath/
+    # The jsonpath root is the objects: element of the template
+    # XXX There is currently an issue with annotations not supporting
+    #     list values
+    template.cnv.io/replace:
+    - path: "/VM/disks/*/type"
+      value: virtio
```
fabiand: This particular example is redundant to the default disk bus. Is there another example that could be used?

templates/README.md
```diff
@@ -0,0 +1,276 @@
...
+    # A single selected template only means no conflicts and no smart
+    # merging code. This has to be done using labels to allow efficient
+    # searching.
+    # The format has the following meaning:
+    # {os,flavor,size}.template.cnv.io/{value}: true (or false for exclusion)
+    # OS names should match the libosinfo identifiers
+    # flavors are tiny, medium, large, etc.
+    # workloads are generic, high-performance, io-intensive,
+    #               oracle-db, sap-hana...
+    os.template.cnv.io/windows2k12r2: "true"
```
fabiand: The scheme does not work.

It works to use boolean to flag a template if it is relevant for an os, workload, size.

What does not work is that the triple is always depending on each other, thus we need to say, this template is applicable to "large windows compute", but from the current mechanism it is not clear if this template applies to "large windows io-intensive" and "large windows minilal" or just one of them.

Instead we could encode the triples in flags:

```
guest.cnv.io/large-windows7-compute
guest.cnv.io/large-windows7-generic
```

templates/fedora28.yaml
```diff
@@ -64,16 +68,17 @@ objects:
             claimName: ${PVCNAME}
         - cloudInitNoCloud:
             userData: |-
-              #cloud-config
```
fabiand: This is a magic header and needs to be kept.

templates/README.md
MarSik: No, it might contain anything that why it is a reference to a template disk.

fabiand:
What is a tempalte disk?

How would it look?

templates/README.md
MarSik: Fabian, focus. Cache was an example and those are references to template specifications so the values are not part of any annotation, but of the disk/volume/interface/network definitions in the spec. That way we can express anything we need.

templates/README.md
MarSik: No, again, read the explanation around line 39.

templates/README.md
MarSik: Nothing for the stock templates (or maybe the L2 request), but the user can provide his own and there.. expected bandwith? network name?

templates/README.md
MarSik: (os1 OR os2 OR os3) AND (workload1 OR workload2) AND (flavor1 OR flavor2) AND NOT anything that is false

And I am not going to express the full boolean logic there as I do not see the need for it. We can use the proposed negative (false) for exceptions and the rest will be implemented using separate template files.

templates/README.md
fabiand: Then let's please make the example clearer to understand what kind of value will be used.

The example looks like a single value, but your words sound like it will contain yaml.

templates/README.md
fabiand: Then please also update this example to provide a meaningful example.

templates/README.md
fabiand: Could you please explain where you would put this logic?

templates/README.md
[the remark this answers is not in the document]
jelkosz: I meant "vm" or "config" as two different values which that label can have. But can have "base" or "config", both work for me. Make your choice :)

templates/LABELS.md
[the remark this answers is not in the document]
jelkosz: great, and than I will have to hope (okey, make sure somehow) that the version of osinfo from which this templates have been generated is compatible with the version of osinfo Im exposing as a service (or compiling the json objects from or whatever solution will have to be used).
I think having it here will save us lots of headache especially during updates.

MarSik:
> What is a tempalte disk? How would it look?

Open any proposed template and you will see it. It is a standard disk, the annotation just references it by name and the UI will use it as example to add disks.

I guess GitHub does not make it easy now to actually see the file as a whole, but it is all explained in the templates.. see line 39 of the README and then for example line 151 in the same file.

templates/README.md
MarSik: Label based object search is already implemented in kubernetes. And searching for a set of labels will be done by the user via the UI. The fact we will organize the templates to follow the described logic does not require any implementation. It will just be done that way when we ship the templates.

templates/README.md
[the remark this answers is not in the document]
MarSik: I would prefer "base" then.

templates/LABELS.md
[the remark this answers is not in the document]
MarSik: The OS names will not change so we should be fairly safe there. We can also ship a proper version of libosinfo with the product, so this should be fine and safe.

templates/README.md
MarSik: Again. Read line 39 which is just above this whole section. And the example here references default-nic from line 167 so it is a proper and working example.

templates/README.md
MarSik: No they don't. Read line 39 first as that explains the meaning of all four defaults together. Pay attention to:

# The information encoded there informs the tooling (UI)
# about names of devices that are to be used as templates
# for adding additional disk, networks, volumes and such

templates/README.md
fabiand: So I read lien 39. To me this is not very well defined. It's unclear how the value should look.
If we don#t define how the value should look, then I do not udnerstand why we should define the label.

templates/README.md
fabiand: Rigth. I understand that in Kubernets you can do boolean expressions.

My point is: By havnig the labels for size, os, and workload independently on a VM, we don#t knwo what triple is support.
This means that all possible permutations of all labels on a VM are supported.
If this is th goal then good, but to me it looks greedy.

MarSik: Yes it is greedy by design.

templates/README.md
MarSik: The value is the device name that is supposed to be used as a template. All the example values are actually pointing to existing disks / volumes /.. with the same name as used in the annotation.

templates/README.md
```diff
@@ -0,0 +1,260 @@
...
+apiVersion: kubevirt.io/v1alpha2
+kind: VirtualMachine
+metadata:
+  name: windows-10-1
+  annotations:
+    # Data used for the UI, ignored by kubevirt
+    # Arbitrary format as needed to be able to
+    # repopulate the UI or the template processor
+    # and get the same output
+    template.cnv.io/original:
```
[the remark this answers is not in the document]
MarSik: Actually, I need to change this section, because it won't work this way. The value has to be a string. I will have to change it to something like parameter.template.cnv.io/{param-name}: value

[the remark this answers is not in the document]
fabiand: I'd actually use a single annotation but then serialize it in json.This will be more predictive, as we know we only have one annotation.

michalskrivanek:
yay!

PR #17, opened 2018-09-13
Travis-CI deploy github releases

ksimon1:
@fabiand 
Added deploy part to travis config, which creates a tarball with all templates/presets and uploads it to github releases.

https://trello.com/c/CURANTGC/26-8-presets-and-profiles-figure-out-upstream-build-and-test-procedures-for-presets

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,18 +14,39 @@ env:
     - CPLATFORM=minikube CDIST=kubernetes CVER=1.11.2
 
 
-matrix:
+jobs:
   include:
-  - name: Check syntax of all templates
-    env: TARGET=syntax-tests
-  - name: Functional test of the Fedora template
-    env: TARGET=functional-tests TEST_FUNCTIONAL=fedora28
-  - name: Functional test of the Ubuntu template
-    env: TARGET=functional-tests TEST_FUNCTIONAL=ubuntu1804
-#  - name: Functional test of the OpenSUSE template
-#    env: TARGET=functional-tests TEST_FUNCTIONAL=opensuse15
-  - name: Functional test of the CentOS/RHEL template
-    env: TARGET=functional-tests TEST_FUNCTIONAL=rhel75
+    - stage: "Tests"
+      name: Check syntax of all templates
+      env: TARGET=syntax-tests
+      # -e is important to let make use the vars defined in "matrix:" above
+      script: bash -x test.sh
+    - script: bash -x test.sh
+      name: Functional test of the Fedora template
+      env: TARGET=functional-tests TEST_FUNCTIONAL=fedora28
+    - script: bash -x test.sh
+      name: Functional test of the Ubuntu template
+      env: TARGET=functional-tests TEST_FUNCTIONAL=ubuntu1804
+#    - script: bash -x test.sh
+#      name: Functional test of the OpenSUSE template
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,20 +9,40 @@ env:
...
+      env: TARGET=functional-tests TEST_FUNCTIONAL=ubuntu1804
+    - script: make -e $TARGET
+      name: Functional test of the CentOS/RHEL template
+      env: TARGET=functional-tests TEST_FUNCTIONAL=rhel75
+    - stage: "Deploy"
+      name: deploy 
+      before_script: skip
+      script: skip
+      before_deploy:
+      - cat presets/*.yaml templates/*.yaml > ./all.yaml
```
fabiand: This does not work, as the ymals need to be divided by `---`.

`make common-templates.yaml` can now do the job: https://github.com/fabiand/common-templates/blob/master/makefile#L30

.travis.yml
```diff
@@ -9,20 +9,40 @@ env:
...
+      - tar -cvf release.tar all.yaml
+      deploy:
```
fabiand: +1 on the deployment.

.travis.yml
```diff
@@ -9,20 +9,40 @@ env:
...
+        provider: releases
+        api_key: $GITHUB_OAUTH_TOKEN
+        file: "release.tar"
```
fabiand: Let's publish the plain yaml. That is much easier to fetch using curl.

For tarballs I'd recommedn to link to the provided tarball, i.e.: https://github.com/fabiand/common-templates/archive/v0.1.0.tar.gz

.travis.yml
```diff
@@ -9,20 +9,40 @@ env:
...
-    env: TARGET=syntax-tests
-  - name: Functional test of the Fedora template
-    env: TARGET=functional-tests TEST_FUNCTIONAL=fedora28
-  - name: Functional test of the Ubuntu template
-    env: TARGET=functional-tests TEST_FUNCTIONAL=ubuntu1804
-#  - name: Functional test of the OpenSUSE template
-#    env: TARGET=functional-tests TEST_FUNCTIONAL=opensuse15
-  - name: Functional test of the CentOS/RHEL template
-    env: TARGET=functional-tests TEST_FUNCTIONAL=rhel75
+    - stage: "Tests"
```
fabiand: Any resaon for these changes?

if possible then I'd like to delay this part.

.travis.yml
ksimon1: fixed

.travis.yml
ksimon1: Reason why i change it is that i didn't find solution how i can add deploy part to this test matrix. If i add deploy part to top level of this config, then it will make deploy in each test. If i add it as a part of this matrix, then it deploys before all tests finish. And per travis documentation, jobs should be used for this case.

.travis.yml
MarSik: That will link to the source though and that will not be good enough once we start compiling the templates.

.travis.yml
MarSik: I agree with Karel here. This seems to be the recommended way per https://docs.travis-ci.com/user/build-stages#build-stages-and-deployments

fabiand: Ok.

PR #26, opened 2018-10-12
Meta templating

MarSik:
The changes in this branch introduce a matrix of templates that are generated using Ansible together with libosinfo data. The reason I chose Ansible is that it supports constants, declarative loops, Jinja2 templates, lookup plugins (libosinfo) and general file manipulation and all that is pretty easy to read.


makefile
```diff
@@ -1,33 +1,35 @@
...
 # i.e. fedora28
-ALL_GUESTS=$(ALL_TEMPLATES:templates/%.yaml=%)
+ALL_GUESTS=$(ALL_TEMPLATES:templates/%.dist.yaml=%)
 
 
 TEST_SYNTAX=$(ALL_GUESTS)
 TEST_UNIT=$(ALL_GUESTS)
 ifeq ($(TEST_FUNCTIONAL),ALL)
-TEST_FUNCTIONAL=fedora28 ubuntu1804 opensuse15 rhel75
+TEST_FUNCTIONAL=fedora-generic-small ubuntu-generic-small opensuse-generic-small rhel7-generic-small centos7-generic-small
```
fabiand: Here you should only list guests which can be tested. Which we know which works.

makefile
```diff
@@ -1,33 +1,35 @@
...
+unit-tests: generate is-deployed
 unit-tests: $(TEST_UNIT:%=%-apply-and-remove)
 unit-tests: $(TEST_UNIT:%=%-generated-name-apply-and-remove)
 
-functional-tests: is-deployed
+functional-tests: generate is-deployed
 functional-tests: $(TEST_FUNCTIONAL:%=%-start-wait-for-systemd-and-stop)
 
-common-templates.yaml: $(SOURCES)
+common-templates.yaml: generate $(SOURCES)
```
fabiand: For this and the trhee above:

The rule should depend on the SOURCES, and the sources should depend on the generate, as the SOURCES are build from the mETASOURCES

README.md
```diff
@@ -1,64 +1,80 @@
...
+$ git clone https://github.com/kubevirt/common-templates
+
+# Pull all submodules
+$ git submodule init update
 
-$ oc process --local -f https://git.io/fNp4Z
-The Template "win2k12r2" is invalid: template.parameters[1]: Required value:
+# Generate the template matrix
+$ pushd common-templates/templates
+$ ansible-playbook generate.yaml
```
fabiand: Maybe you can rename generate.yaml to something more specific.

I.e. generate-templates.yaml

makefile
```diff
@@ -1,33 +1,35 @@
 SHELL=/bin/bash
 
 # i.e. fedora28.yaml
-ALL_TEMPLATES=$(wildcard templates/*.yaml)
+ALL_META_TEMPLATES=$(wildcard templates/templates/*.yaml)
```
fabiand: Could meta templates be in `tempaltes/meta`?

makefile
```diff
@@ -1,33 +1,35 @@
...
+ALL_TEMPLATES=$(wildcard templates/*.dist.yaml)
```
fabiand: We can have all built artifacts in `build/*` or so, then it's easier to distinguish it from source content.

Having tempaltes in `build/` (or `build/template/*`) would also make the readme nicer, as it would be: `kubectl apply -f build/windows.yaml`

templates/lookup_plugins/osinfo.py
```diff
@@ -0,0 +1,96 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
```
fabiand: Please add a a description to this script.

What does it do, why and what is it used for?

fabiand: Copyright?

makefile
MarSik: dist/ is the common name for such directory, sure I can change that. But then we should probably do a bigger change to the repo. Move the ansible file to the root directory and use the templates dir for the meta templates and dist dir for results.

makefile
MarSik: That is not possible, the generated files do not exist at the beginning and SOURCES is empty.

makefile
MarSik: No, ansible would not find it there. But I removed the nesting in favor of dist/ and templates/

templates/lookup_plugins/osinfo.py
MarSik: Sure, but what license do we want to use?

PR #28, opened 2018-11-01
add iothreads feature

sHaggYcaT:
Add ioThreads feature in the Linux template generator.
By default uses **shared** ioThread policy.  But adds it as editable parameter. 
Current templates uses 2 cores maximum. It's a reason why I choose **shared** ioThread policy.

If user edit .spec.template.spec.domain.cpu.cores, then it can have a wish to edit ioThread policy to. 
So, user has this option.

In this time CNV supports only two policies - _shared_ and _auto_.

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -33,7 +33,7 @@ objects:
     template:
       spec:
         domain:
-{%- if item.iothreads == true %} {{''}}
+{%- if item.iothreads %}
           ioThreadsPolicy: shared
 {% endif %}
 
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -32,7 +32,7 @@ objects:
     template:
       spec:
         domain:
-{%- if item.iothreads %} {{''}}
+{% if item.iothreads %}
           ioThreadsPolicy: shared
 {% endif %}
 
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -32,7 +32,7 @@ objects:
     template:
       spec:
         domain:
-{%- if item.iothreads %}
+{%- if item.iothreads %} {{''}}
           ioThreadsPolicy: shared
 {% endif %}
 
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -12,7 +12,6 @@
       /objects[0].spec.template.spec.domain.devices.disks
       /objects[0].spec.template.spec.volumes
       /objects[0].spec.template.spec.networks
-      /objects[0].spec.template.spec.ioThreadsPolicy
 
   labels:
 {% for osl in oslabels %}
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -12,6 +12,7 @@
       /objects[0].spec.template.spec.domain.devices.disks
       /objects[0].spec.template.spec.volumes
       /objects[0].spec.template.spec.networks
+      /objects[0].spec.template.spec.ioThreadsPolicy
 
   labels:
 {% for osl in oslabels %}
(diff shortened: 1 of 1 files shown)


templates/_linux.yaml
```diff
@@ -32,6 +33,11 @@ objects:
     template:
       spec:
         domain:
+{%- if item.iothreads == true %} {{''}}
+          ioThreadsPolicy: shared
+{%- elif item.iothreads == false or item.iothreads is not defined %}
```
MarSik:
```suggestion
```

You can drop the line, it is not necessary.

templates/_linux.yaml
```diff
@@ -32,6 +33,11 @@ objects:
     template:
       spec:
         domain:
+{%- if item.iothreads == true %} {{''}}
```
MarSik: Why not just?

```suggestion
{%- if item.iothreads %}
```

templates/_linux.yaml
```diff
@@ -12,6 +12,7 @@
       /objects[0].spec.template.spec.domain.devices.disks
       /objects[0].spec.template.spec.volumes
       /objects[0].spec.template.spec.networks
+      /objects[0].spec.template.spec.ioThreadsPolicy
```
MarSik:
```suggestion
```
The global VM policy should stay shared. The user can select diffferent policy per disk (like a dedicated thread for specific disk only).

MarSik:
[(3)[presets-and-profiles] Add ioThreads to high performance templates [initiative-high-performance-vm]](https://trello.com/c/foVvLgUV/65-3presets-and-profiles-add-iothreads-to-high-performance-templates-initiative-high-performance-vm)

templates/_linux.yaml
sHaggYcaT: it's just double check - "never trust data input". In a case if we choose something strange in playbook.

templates/_linux.yaml
sHaggYcaT: it's just double check - "never trust data input". In a case if we choose something strange in playbook.

templates/_linux.yaml
sHaggYcaT: But why Stu wrote this feature (different policy?) Do we use it anywhere?

stu-gott: There are technically 3 modes. "shared", "auto" and "none". The default should probably be to omit the IOThreadsPolicy setting and thus disable it?

The feature @MarSik is referring to is a per-disk override `dedicatedIOThread` that will allocate a thread for a single disk regardless of what the rest of the template is using.

MarSik: That is exactly what I had in mind. Thanks @stu-gott for clarification.

templates/_linux.yaml
MarSik: That is our responsibility, the templates here and the parameters are code we own and maintain. This check specifically will give you no warning anyway.

templates/_linux.yaml
MarSik: But the else section gives you nothing - no error, no warning - and is simply dead code.

templates/_linux.yaml
```diff
@@ -32,6 +32,10 @@ objects:
     template:
       spec:
         domain:
+{%- if item.iothreads %} {{''}}
```
MarSik: Why did you keep the {{''}} part? I do not quite understand its purpose here..

sHaggYcaT: Just run playbook without {{''}} and check - the template will be wrong. {{''}} helps to add additional \n in case if iothreads exist. 

Checked in the ansible 2.7.0

PR #31, opened 2018-11-01
add other templates to test

ksimon1:
add other sizes of templates to tests. Due to travis restrictions, large templates have to be updated to use only 6GB of RAM.


automation/x-limit-ram-size.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Due to travis restrictions, travis job instances have only 7.5 GB of memory.
+# This script updates ram size in all large templates 
+for filename in dist/templates/*-large.yaml; do
```
MarSik: The loop can actually go over all templates I think. It would still affect only the large ones.

PR #32, opened 2018-11-07
fix bug when more tests use same image

ksimon1:
when more tests run, sometimes it takes lot of time to delete previous vmi. And there is an issue, that kubernetes deletes image which new vm needs when it deletes old vm.


MarSik:
[(5)[presets-and-profiles] Functional tests for Windows template [initiative-always-ready-release]](https://trello.com/c/SPjqYB0J/41-5presets-and-profiles-functional-tests-for-windows-template-initiative-always-ready-release)

PR #35, opened 2018-11-12
Build osinfo-db properly and include full OS names in templates

MarSik:


MarSik:
[(3)[templating-1.4] Add full OS name hints to all templates [initiative-vm-ops-ui]](https://trello.com/c/sehJRdQE/81-3templating-14-add-full-os-name-hints-to-all-templates-initiative-vm-ops-ui)

PR #44, opened 2018-12-12
Fix windows test

ksimon1:
This patch updates kubevirt version in windows test to 0.11 (because of KUBEVIRT_MEMORY_SIZE variable), updates windows image download function (now it should not timeout because of download lock), removes restriction for vm memory size (now we can set more memory for node)

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -17,50 +17,59 @@
 # Copyright 2018 Red Hat, Inc.
 #
 
-trap '{ release_download_lock $WINDOWS_LOCK_PATH; make cluster-down; }' EXIT SIGINT SIGTERM SIGSTOP
-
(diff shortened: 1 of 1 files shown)


ksimon1:
@MarSik, @fromanirh, @yanirq please review

generate-templates.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 - connection: local
   hosts: 127.0.0.1
+  gather_facts: no
```
MarSik: This should probably go to a separate patch as it applies to Travis as well. It is correct though.

ksimon1: done - https://github.com/kubevirt/common-templates/pull/45

PR #45, opened 2018-12-14
turn off ansible gather_facts

ksimon1:
This patchs turns off ansible gather_facts, because it fails stdci test


MarSik:
And also because we are not using any local facts :)

PR #46, opened 2018-12-17
Template validation docs

fromanirh:
Add documentation for the validation annotation, to be consumed by the UI tools.

### DIFF
--- a/templates/VALIDATION.md
+++ b/templates/VALIDATION.md
@@ -22,22 +22,24 @@ validations: |
   [
     {
       "name": "validation-rule-01",
-      "valid": "/some/json/path",
-      "path”: "/some/json/path/*/leaf",
+      "valid": "/some.json.path",
+      "path”: "/some.json.path[*].leaf",
       "rule”: "integer",
-      "message”: "/some/json/path/*/leaf must exists",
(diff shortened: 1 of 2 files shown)


templates/VALIDATION.md
```diff
@@ -0,0 +1,220 @@
...
+## Summary
+
+We introduce and document here a new annotation format that serves as hint for CNV-aware tooling to validate the templates.
+Being built on the existing Kubernetes annotation format, the data expressed using this format is completely optional and ignored by existing Kubernetes/Openshift services.
+
+In order to be backward and future compatible, consumers of these annotations should ignore the data they don't understand.
+
+## Format
+
+The validation hints are encoded into another template annotation, called `validation`.
```
rawagner: nitpick `validations` (plural)

templates/VALIDATION.md
```diff
@@ -100,7 +102,7 @@ kind: Template
         [
           {
             “rule”: “integer”,
-            "path": "/VM/domain/cpu/cores",
+            "path": "/spec.domain.cpu.cores",
```
MarSik: If the root is the objects: node then I think this won't work, because the objects: node is an array. It might work without the / though

PR #49, opened 2019-01-11
travis: Use token from env var

fabiand:
Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>


MarSik:
Already handled by different PR.

PR #54, opened 2019-02-12
doc: validation: add clarifications, close gaps

fromanirh:
In this patch:
1. minor fixes for consistency with JSONPath spec:
   https://goessner.net/articles/JsonPath/index.html
2. clarify the missin "regex" argument
3. require notation to disambiguate JSONPath from strings

### DIFF
--- a/templates/VALIDATION.md
+++ b/templates/VALIDATION.md
@@ -22,19 +22,19 @@ validations: |
   [
     {
       "name": "validation-rule-01",
-      "valid": "$.some.json.path",
-      "path”: "$.some.json.path[*].leaf",
+      "valid": ".some.json.path",
+      "path”: ".some.json.path[*].leaf",
       "rule”: "integer",
-      "message”: "$.some.json.path[*].leaf must exists",
+      "message”: ".some.json.path[*].leaf must exists",
       "min”: 1
     },
     {
       "name": "validation-rule-02",
-      "valid": "$.another.json.path",
-      "path": "$.another.json.path.item",
+      "valid": ".another.json.path",
+      "path": ".another.json.path.item",
       "rule": "integer",
-      "message": "$./another.json.path.item must be below a threshold",
-      "max": "$.yet.another.json.path.defines.the.limit"
+      "message": "./another.json.path.item must be below a threshold",
+      "max": ".yet.another.json.path.defines.the.limit"
     }
   ]
 ```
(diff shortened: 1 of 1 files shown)


templates/VALIDATION.md
```diff
@@ -51,6 +51,18 @@ If a rule is meaningless, for example if it has no arguments (see below), its be
 ### JSONPaths
 
 For every jsonpath mentioned in this document, unless specified otherwise, the root is the objects: element of the template.
+Unless otherwise specified, the jsonpath must be enclosed in double brackets - this is to disambiguate plain strings from jsonpaths.
+Example:
+
+good:
+```
+{{.spec.domain.resources.requests.memory}}
```
MarSik: This is problematic. Double brackets are used in ansible and we use ansible to generate the final template files. We would create escaping hell this way..

templates/VALIDATION.md
```diff
@@ -22,19 +22,19 @@ validations: |
...
       "name": "validation-rule-02",
-      "valid": "/another.json.path",
-      "path": "/another.json.path.item",
+      "valid": "$.another.json.path",
+      "path": "$.another.json.path.item",
       "rule": "integer",
-      "message": "/another.json.path.item must be below a threshold",
-      "max": "/yet.another.json.path.defines.the.limit"
+      "message": "$./another.json.path.item must be below a threshold",
+      "max": "$.yet.another.json.path.defines.the.limit"
```
MarSik: You use `$.` here and `.` below. Care to explain the difference?

fromanirh: It should be $. everywhere for consistency, even though both are legal per https://goessner.net/articles/JsonPath/index.html#e3

templates/VALIDATION.md
fromanirh: No big deal, I just need a clean and simple way to dfistinguish what is supposed to be a jsonpath and what must be treated as string literal. Any syntax is fine. Maybe

  jsonpath::.spec.domain.resources.requests.memory

? aka "jsonpath::" prefix

templates/VALIDATION.md
```diff
@@ -221,6 +233,28 @@ kind: Template
...
+kind: Template
+  metadata:
+    name: windows-10
+    annotations:
+      validations: |
+        [
+          {
+            "name": "supported-bus",
+            "path": ".spec.devices.disks[*].type",
+            "rule": "enum",
```
fromanirh: obv. "rule": "regex",

templates/VALIDATION.md
```diff
@@ -51,12 +51,12 @@ If a rule is meaningless, for example if it has no arguments (see below), its be
 ### JSONPaths
 
 For every jsonpath mentioned in this document, unless specified otherwise, the root is the objects: element of the template.
-Unless otherwise specified, the jsonpath must be enclosed in double brackets - this is to disambiguate plain strings from jsonpaths.
-Example:
+Unless otherwise specified, the value to be used as jsonpath must be prefixed with the "jsonpath::" literal.
```
MarSik: This confused me a bit when I looked at the updated example above. I think we should make it clear that this only applies to the value field. All other fields like valid or path do not need this prefix.

PR #55, opened 2019-02-28
removed generic workload

ksimon1:
removed generic workload - replaced by desktop/server workloads
updated tests
@MarSik please review


ksimon1:
@MarSik please review

PR #61, opened 2019-03-06
Fix #60 - template api version

### LINKED ISSUE #60: Adjust apiVersion to have  the openshift path
Currently the [ssp operator](https://github.com/MarSik/kubevirt-ssp-operator), an ansible operator based on [openshift-enterprise-ansible-operator](http://pkgs.devel.redhat.com/cgit/containers/openshift-enterprise-ansible-operator/), cannot identify the templates without having a full api version in the templates.

To add the compatibility, all apiVersion fields in the templates should change to:
`apiVersion: template.openshift.io/v1`

MarSik:
Fixes https://github.com/kubevirt/common-templates/issues/60

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
   name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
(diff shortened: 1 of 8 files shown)


yanirq:
@MarSik  is the output file being produced here will include only the supported templates ?

MarSik:
One test had to be restarted, but all are passing.

PR #62, opened 2019-03-14
Several additions to base Linux template

kjw3:
- Add single default pod networking nic
- Add metadata for template so expose port in virtctl works
- Add cloudinit disk so cloud init works

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -38,6 +38,10 @@ objects:
   spec:
     running: false
     template:
+      metadata:
+        labels:
+          kubevirt.io/domain: ${NAME}
+          kubevirt.io/size: {{ item.flavor }}
       spec:
         domain:
 {% if item.iothreads %}
@@ -66,7 +70,16 @@ objects:
             - disk:
                 bus: {{ diskbus | default("virtio") }}
               name: rootdisk
+            - disk:
+                bus: {{ diskbus | default("virtio") }}
+              name: cloudinitdisk
+            interfaces:
+            - bridge: {}
+              name: default
         terminationGracePeriodSeconds: 0
+        networks:
+        - name: default
+          pod: {}
         volumes:
         - name: rootdisk
           persistentVolumeClaim:
@@ -76,7 +89,7 @@ objects:
               #cloud-config
               password: {{ password }}
               chpasswd: { expire: False }
-          name: cloudinitvolume
+          name: cloudinitdisk
 
 parameters:
 - description: VM name


ovirt-infra:
Hello contributor, thanks for submitting a PR for this project! 

I am the bot who triggers "standard-CI" builds for this project.
As a security measure, I will not run automated tests on PRs that are not from white-listed contributors.

In order to allow automated tests to run, please ask one of the project maintainers to review the code and then do one of the following:

1. Type `ci test please` on this PR to trigger automated tests for it.
2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically. ( keep in mind this list might be overwritten if the job XML is refreshed, for permanent whitelisting, please follow #3 option )
3. If you are planning to contribute to more than one project, maybe it's better to ask them to add you to the project organization, so you'll be able to run tests for all the organization's projects.

templates/_linux.yaml
```diff
@@ -38,6 +38,10 @@ objects:
   spec:
     running: false
     template:
+      metadata:
+        labels:
+          kubevirt.io/domain: ${NAME}
```
MarSik: What will happen to VMs created by hand? I would expect this specific label to be added by the modifying webhook in kubevirt when VMI is created.

templates/_linux.yaml
```diff
@@ -66,7 +70,16 @@ objects:
             - disk:
                 bus: {{ diskbus | default("virtio") }}
               name: rootdisk
+            - disk:
```
MarSik: Wasn't this disk added by default by kubevirt core components? But the truth is I never noticed we have volume and no disk...

templates/_linux.yaml
```diff
@@ -66,7 +70,16 @@ objects:
...
+                bus: {{ diskbus | default("virtio") }}
+              name: cloudinitdisk
+            interfaces:
+            - bridge: {}
+              name: default
         terminationGracePeriodSeconds: 0
+        networks:
```
MarSik: I am not sure this should be added by default. There are cases where a VM without networking might be needed.

templates/_linux.yaml
kjw3: I don't know the answer to your question. I just know without labels in metadata the virtclt expose command won't work for the vmi

templates/_linux.yaml
kjw3: The disk does not get add... it doesn't complain about it not being there. The machine just hangs for a long time at boot and if this disk is added to the template the machine boots as expected and cloud-init is processed.

templates/_linux.yaml
kjw3: All of the base cloud images I've ever seen expect the VM to have one nic. The RHEL and CentOS 7 images as example have a dhcp ifcfg for eth0

kjw3:
You can obvioulsy modify this as needed, I just expected cloud-init and a single nic and the expose command to work on the VM by default as a result of using one of these common templates.

I'll leave it to you all to make best decision about what should be or should not be included.

MarSik:
We can include this and it sounds right indeed. I just wanted to know the why part of the story.
But a question about the kubevirt label should be opened in kubevirt too, I would expect they add it to every VMI by default.

PR #66, opened 2019-05-10
Add required memory validation rules (and update the osinfodb)

MarSik:
This patch is an example of how to add validations that are based on libosinfo data. So far only the minimal memory requirements are checked.


MarSik:
@ksimon1 This might be good enough to add minimal memory requirements to guest templates.

ksimon1:
@MarSik i think you can close this PR

MarSik:
Closed in favor of #69

PR #65, opened 2019-04-15
increased kubevirt version to v0.18.0

ksimon1:

PR #67, opened 2019-05-27
linux: windows: set evictionStrategy explicitely

fromanirh:
Bug-Url: https://bugzilla.redhat.com/1714000

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ branches:
   - /^v[0-9]/
 env:
   global:
-  - KUBEVIRT_VER=v0.15.0-alpha.0
+  - KUBEVIRT_VER=v0.17.0
   - CPLATFORM=minikube CVER=1.11.2
 jobs:
   include:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ branches:
   - /^v[0-9]/
 env:
   global:
-  - KUBEVIRT_VER=v0.15.0-alpha.0
+  - KUBEVIRT_VER=v0.17.0
```
ksimon1: PR is already created and waiting for merge https://github.com/kubevirt/common-templates/pull/65

MarSik:
I do not think we want to merge this. We decided in the past that upstream should not enable live migration by default.

MarSik:
Also the root cause was elsewhere - https://bugzilla.redhat.com/show_bug.cgi?id=1711457

fromanirh:
Thanks @MarSik . I missed the past decision, and I thought we needed *both* the validator and the common templates fix. But looks like the validator fix could be sufficient, so I'm closing this PR.

PR #68, opened 2019-05-28
move rhel8 test from travis to stdci

ksimon1:
This PR moves RHEL 8 test from travis to stdci. For now, rhel 8 image is not presented on stdci server, so test is using rhel 7 image.


automation/check-patch.mounts
```diff
@@ -2,4 +2,4 @@
 /var/lib/libvirt/boot
 /var/cache:/var/host_cache
 /var/run/docker.sock:/var/run/docker.sock
-/var/lib/stdci/shared:/var/lib/stdci/shared
+/var/lib/stdci/shared:/var/lib/stdci/shared
```
fromanirh: spurious change? the two lines look identical. No big deal however

automation/check-patch.rhel8.packages
```diff
@@ -0,0 +1,7 @@
+git
+docker
+jq
+ansible
+libosinfo
+expect
+python-gobject
```
fromanirh: what is python-gojbect needed for?

automation/check-patch.windows2016.packages
```diff
@@ -3,4 +3,4 @@ docker
 jq
 ansible
 libosinfo
-python-gobject
+python-gobject
```
fromanirh: same, looks spurious change

automation/check-patch.mounts
fromanirh: ...but as guiding principle, we should minimize patch/changesets/PRs.

automation/test-rhel8.sh
```diff
@@ -0,0 +1,117 @@
...
+    jq '.items[0].spec.template.spec.volumes[0]+= {"ephemeral": {"persistentVolumeClaim": {"claimName": "disk-rhel"}}} | 
+    del(.items[0].spec.template.spec.volumes[0].persistentVolumeClaim)' | \
+    _oc apply -f -
+
+    # start vm
+    ./virtctl --kubeconfig=$kubeconfig start "$template_name-$workload-$size"
+
+    set +e
+    current_time=0
+    while [[ $(_oc get vmi $template_name-$workload-$size -o json | jq '.status.phase') != *Running* ]] ; do
```
fromanirh: do you need some quotation marks, e..g "" around *Running* ? Ypu may want to use the -r option of jq.

automation/test-rhel8.sh
```diff
@@ -0,0 +1,117 @@
...
+    set -e
+
+
+    # get ip address of vm
+    ipAddressVMI=$(_oc get vmi $template_name-$workload-$size -o yaml | grep ipAddress | awk '{print $3}')
+
+    set +e
+    current_time=0
+    # Make sure vm is ready
+    while _oc exec -it winrmcli -- ping -c1 $ipAddressVMI| grep "Destination Host Unreachable" ; do
```
fromanirh: do we need winrmcli for RHEL8 tests?

automation/test-windows.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
```
fromanirh: could you please elaborate why do we need this file in the RHEL8 PR?

automation/check-patch.rhel8.packages
ksimon1: I think it is needed for communication between osinfo-db and script which creates templates.

automation/test-rhel8.sh
ksimon1: this container is needed to check if VM is online or not (it pings pod with VM). Of course we can use e.g. busybox or something similar.

automation/test-windows.sh
ksimon1: this file was created, because previously we had only one test for windows. With new rhel test i needed to separate those two tests.

fromanirh:
@ksimon1 my last question is about using winrmcli, otherwise LGTM. If CI keep working, we should be fine.

ksimon1:
winrmcli is part of the kubevirt project (it has image under kubevirt organization on dockerhub)

fromanirh:
> winrmcli is part of the kubevirt project (it has image under kubevirt organization on dockerhub)

OK, se let's go ahead. LGTM.

automation/test-rhel8.sh
MarSik: I am fine with this, but it looks weird to see winrmcli in rhel8 test :) I understand we are using it just because it already includes the necessary tools.

automation/test.sh
```diff
@@ -19,10 +19,30 @@
...
+readonly TEMPLATES_SERVER="https://templates.ovirt.org/kubevirt/"
+
+export RHEL_NFS_DIR=${RHEL_NFS_DIR:-/var/lib/stdci/shared/kubevirt-images/rhel7}
+export RHEL_LOCK_PATH=${RHEL_LOCK_PATH:-/var/lib/stdci/shared/download_rhel_image.lock}
 export WINDOWS_NFS_DIR=${WINDOWS_NFS_DIR:-/var/lib/stdci/shared/kubevirt-images/windows2016}
 export WINDOWS_LOCK_PATH=${WINDOWS_LOCK_PATH:-/var/lib/stdci/shared/download_windows_image.lock}
-export KUBEVIRT_MEMORY_SIZE=12288M
+export KUBEVIRT_MEMORY_SIZE=16384M
+export KUBEVIRT_PROVIDER="os-3.11.0"
+export VERSION="v0.18.0"
```
MarSik: I wonder if we should extract this somewhere to make it easier to maintain and update. Could we autodetect the latest for example?

ksimon1: Good idea, it should not be so hard to do it

fromanirh:
> @fromanirh Could you help Karel here? I can't find the code where we discussed this recently.

@MarSik sure thing. I think you're referring to https://github.com/MarSik/kubevirt-ssp-operator/pull/57

PR #69, opened 2019-06-17
Add required memory validation rules (and update the osinfodb)

ksimon1:
This patch is an example of how to add validations that are based on libosinfo data. So far only the minimal memory requirements are checked.
Updated version of osinfo-db

### DIFF
--- a/lookup_plugins/osinfo.py
+++ b/lookup_plugins/osinfo.py
@@ -61,7 +61,8 @@ class OsInfoGObjectProxy(object):
             return False
 
     def _resolve(self, val, path):
-        if (type(val) == int or type(val) == float or type(val) == str or
+        if (type(val) == int or type(val) == long or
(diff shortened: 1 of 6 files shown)


MarSik:
Hmm, but the long check is needed for Python 2 and large numbers. I wonder if we should use 

```
import six
...
if type(val) in six.integer_types
```

ksimon1:

lookup_plugins/osinfo.py
```diff
@@ -61,7 +61,8 @@ def __contains__(self, key):
             return False
 
     def _resolve(self, val, path):
-        if (type(val) == int or type(val) == float or type(val) == str or
+        if (type(val) == int or type(val) == long or
+                type(val) == float or type(val) == str or
```
fromanirh: I don't like type checks entirely, but this is stuff for another OR.

PR #70, opened 2019-06-20
windows: templates: use synic flags

fromanirh:
Kubevirt gained support for pretty much all the hyper-v related flags
exposed in libvirt 5.1.
In this patch we update the common templates to use some of them.
We start with the following:

- synic: needed to enable synictimer
- synictimer: reportedly improves performance significantly
     on windows >= 10
- vpindex: needed for few optimization on overcommitted environments
     (ipi, tlbflush). It is expected NOT to harm performances if enabled
     alone, we enable to prepare for the aforementioned optimizations.

Please note that we DO NOT assume or require the HyperV strict
checking[1].
We leave that to the cluster admin or to some other opinionated setups.

+++

[1]: https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/guest-operating-system-information.html#hyperv-optimizations

### DIFF
--- a/automation/test-windows.sh
+++ b/automation/test-windows.sh
@@ -134,6 +134,7 @@ for size in ${sizes[@]}; do
         exit 1
       fi
       sleep 10;
+      _oc exec -it winrmcli -- ./usr/bin/winrm-cli -hostname $ipAddressVMI -port 5985 -username "Administrator" -password "Heslo123" "ipconfig /all"
     done
 
     ./virtctl --kubeconfig=$kubeconfig stop $template_name-$workload-$size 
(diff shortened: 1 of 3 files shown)


fromanirh:
The PR fails the validation because the kernel doesn't allow/support the setting of the vpindex flag, exactly what "HypervStrictCheck" (see https://kubevirt.io/user-guide/docs/latest/creating-virtual-machines/guest-operating-system-information.html#hyperv-optimizations) is supposed to prevent.
Thus, we should first update our CI settings before to be able to verify this PR.

fromanirh:
A good first step to update CI would be merge this PR first https://github.com/kubevirt/common-templates/pull/68 and fill any remaining gaps.

fromanirh:
as it stands today, we can't merge this PR until we figure out why CI breaks.

fromanirh:
abandoned - can't work on this anymore

PR #74, opened 2019-06-27
templates: add versioning information

fromanirh:
We need to know from which version of a template a VM was baked.

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -34,6 +34,7 @@ objects:
     name: ${NAME}
     labels:
       vm.kubevirt.io/template: {{ os }}-{{ item.workload }}-{{ item.flavor }}
+      vm.kubevirt.io/template.version: "{{ lookup('env', 'VERSION') | regex_replace('^v?(?P<version>\d+\.\d+\.\d+)$', '\\g<version>') }}"
       app: ${NAME}
   spec:
     running: false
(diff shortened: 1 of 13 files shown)


templates/_linux.yaml
```diff
@@ -34,6 +34,7 @@ objects:
     name: ${NAME}
     labels:
       vm.kubevirt.io/template: {{ os }}-{{ item.workload }}-{{ item.flavor }}
+      vm.kubevirt.io/template.version: "{{ lookup('env', 'VERSION') | regex_replace('^v?(?P<version>\d+\.\d+\.\d+)$', '\\g<version>') }}"
```
MarSik: There can only be one template with the same name so maybe we should expose the version in the template name directly. In other words, why not create NVR naming scheme?

fromanirh: good idea, but not sure how's gonna play out with non-tag versions. Let's say we do scratch builds. What `version` should be?
Furthermore, can we ever have Revision != 1 ?

MarSik: Upstream we will definitely always have revision = 1. For scratch builds.. I need to think about this, because I do not want to install new set of templates every time a developer does a commit. Maybe we could encode NV in name and encode the R in annotation just for reference purposes.

fromanirh: changing the `name` to `name-version` and encoding the revision into a annotation sounds good to me.
However, this will make upgrade from last released version interesting. We will effectively have duplicate templates, not sure how we can handle that

MarSik: That is the idea. I wanted to allow something like compatibility levels using template versions. Hmm to be able to search for the right one we need to add a label with version too.

fromanirh: ack, got it

fromanirh:
I think the failure in the rhel7 lane is a false negative. Can we retrigger the test?

PR #77, opened 2019-07-16
added description of justWarning flag into validation documentation

ksimon1:
Related to: https://github.com/fromanirh/kubevirt-template-validator/pull/20
This pr added description of justWarning flag into validation documentation.
/cc @MarSik, @fromanirh

### DIFF
--- a/templates/VALIDATION.md
+++ b/templates/VALIDATION.md
@@ -26,7 +26,8 @@ validations: |
       "path”: "jsonpath::.some.json.path[*].leaf",
       "rule”: "integer",
       "message”: ".some.json.path[*].leaf must exists",
-      "min”: 1
+      "min”: 1,
+      "justWarning": true
     },
     {
       "name": "validation-rule-02",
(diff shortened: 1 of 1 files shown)


templates/VALIDATION.md
```diff
@@ -148,6 +149,7 @@ kind: Template
 * `valid`: the rule must be *ignored* if the jsonpath given as value doesn't exist.
 *PLEASE NOTE* that even if values of this key are required to be JSONPaths, you still need to use the `jsonpath::` prefix
 as explained above.
+* `justWarning`: if applying of the rule with this flag fails, it will be only logged into validator logs and validation will not fail.
```
MarSik: "violating rule with justWarning field set will emit a warning only instead of failing the validation"

templates/VALIDATION.md
```diff
@@ -258,6 +260,29 @@ kind: Template
         ]
 ```
 
+### justWarning:
+When applying of the rule is not satisfied and `justWarning` is set, then validator will log that rule failed, but validation will
```
MarSik: The validation normally fails when a rule is not satisfied. The behaviour can be changed per rule by setting the `justWarning` property of the rule. The validator will then emit a warning only and the overall result of the validation will be unaffected by the rule.

ksimon1:

PR #78, opened 2019-07-17
win templates: add warning virtio bus validation

ksimon1:
This PR adds warning validation to windows templates. It checks if vm has virtio disk bus.
/cc @MarSik, @fromanirh

### DIFF
--- a/templates/win2k12r2.tpl.yaml
+++ b/templates/win2k12r2.tpl.yaml
@@ -39,6 +39,13 @@ metadata:
           "rule": "integer",
           "message": "This VM requires more memory.",
           "min": {{ lookup('osinfo', osinfoname)["minimum_resources.0.ram"] }}
+        }, {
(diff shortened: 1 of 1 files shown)


templates/win2k12r2.tpl.yaml
```diff
@@ -39,6 +39,13 @@ metadata:
           "rule": "integer",
           "message": "This VM requires more memory.",
           "min": {{ lookup('osinfo', osinfoname)["minimum_resources.0.ram"] }}
+        }, {
+          "name": "windows-virtio-bus",
+          "path": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
```
MarSik: What about exposing two rules? One enforcing that tells the user that valid options are virtio and sata and the other like you wrote - just warning.

ksimon1:

PR #79, opened 2019-07-19
templates: fix windows test

ksimon1:
this PR fixes check if win vms are up, adds retry logic, disables win2k19 test

### DIFF
--- a/automation/test-windows.sh
+++ b/automation/test-windows.sh
@@ -66,7 +66,7 @@ restartPolicy: Always
 ---
 EOF
 
-timeout=600
+timeout=400
 sample=30
 
 # Make sure winrmcli pod is ready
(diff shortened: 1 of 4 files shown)


automation/test-windows.sh
```diff
@@ -123,7 +123,7 @@ for size in ${sizes[@]}; do
     set +e
     current_time=0
     # run ipconfig /all command on windows vm
-    while [[ $(_oc exec -it winrmcli -- ./usr/bin/winrm-cli -hostname $ipAddressVMI -port 5985 -username "Administrator" -password "Heslo123" "ipconfig /all") != *"$ipAddressVMI"* ]] ; do 
+    while [[ $($(_oc exec -it winrmcli -- ./usr/bin/winrm-cli -hostname $ipAddressVMI -port 5985 -username "Administrator" -password "Heslo123" "ipconfig /all") | grep "IPv4 Address") != *"$ipAddressVMI"* ]] ; do
```
fromanirh: I'm not sure you need the double shell exec (nested $()) here. Couldn't you just use a pipe?

MarSik: The nesting is actually wrong, it will first generate new command "output of winrm | grep" and execute it

ksimon1: fixed

fromanirh:
Disabling win2019 test is unfortunate, but since the failures seems to be related only to that lane, we can't stop the world while we troubleshoot the issue, especially since we are looking to add libvirt/qemu settings recommended from long time (aka not experimental or risky)

fromanirh:
this improves CI, so I think we should have it in.

PR #96, opened 2019-10-17
Refactoring of travis tests

ksimon1:
travis tests are now in separate bash files.
removed unnecessary commands in makefile

Signed-off-by: ksimon1 <ksimon@redhat.com>

### DIFF
--- a/travis_ci/test.sh
+++ b/travis_ci/test.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2019 Red Hat, Inc.
+#
+set -ex
+if [ -z "$1" ]; then
+	echo "First parameter should contain name of OS" && exit 1
+fi
+
+if [ -z "$2" ]; then
+	echo "Second parameter should contain workload of OS" && exit 1
+fi
+
+name=$1
+workload=$2
+
+travis_fold_start(){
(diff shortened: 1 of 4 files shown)


travis_ci/test.sh
```diff
@@ -0,0 +1,104 @@
...
+mkdir -p "$PWD/pvs/$name"
+qemu-img convert -p -O raw $name "$PWD/pvs/$name/disk.img"
+sudo chown 107:107 "$PWD/pvs/$name/disk.img"
+sudo chmod -R a+X "$PWD/pvs"
+
+size_MB=$(( $(qemu-img info $name --output json | jq '.["virtual-size"]') / 1024 / 1024 + 128 ))
+# Create PV and PVC
+bash create-minikube-pvc.sh "$name" "${size_MB}M" "$PWD/pvs/$name/" | tee | oc apply -f -
+
+sizes=("tiny" "small" "medium" "large")
```
MarSik: Not all OSes have to have all sizes. We moved the complicated ones to stdci though (Windows).

ksimon1:
cc @MarSik, @fromanirh

travis_ci/test.sh
```diff
@@ -0,0 +1,111 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2019 Red Hat, Inc.
+#
+
```
fromanirh: please add `set -e` right at the beginning, so the script bail out at the first error. Do we want also to add `-x`, so the shell is automatically print out the command it is about to run?

travis_ci/test.sh
```diff
@@ -0,0 +1,111 @@
...
+name=$1
+workload=$2
```
fromanirh: please add code to handle missing parameters, it should print a short help/usage message and the exit

travis_ci/test_syntax.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2019 Red Hat, Inc.
+#
+
```
fromanirh: same comment about  `set -e`

PR #97, opened 2019-10-22
Released templates are automaticaly pushed to ssp operator.

ksimon1:
When new release of common templates is created, travis will push new file and update default version of common templates in ssp-operator.

Do NOT merge this PR. It contains wrong keys and variable names. It is just a POC to get some feedback.

/cc @MarSik, @fromanirh  

Signed-off-by: ksimon1 <ksimon@redhat.com>


MarSik:
@fromanirh The script is directly from the Travis docs so there is nothing more we can do. It all runs in an ephemeral container. Travis has the decryption keys anyway so we have to trust it.

Unfortunately there is no way to give it PR permissions only, it needs to push somewhere first.

fromanirh:
@MarSik ack

PR #102, opened 2019-10-31
Fix deployment key encryption to use with travis-ci.com

MarSik:
Turns out it is not enough to be logged in to travis --com, the encrypt command needs the --com argument also.


MarSik:
See https://travis-ci.com/kubevirt/common-templates/settings

PR #103, opened 2019-11-06
Fix examples on the README page

ahadas:
Signed-off-by: Arik Hadas <ahadas@redhat.com>


fromanirh:
Fixing the README can hardly broke the build, so I'd waive the failure. But first, let me see what actually is, we could have catched a real issue by accident.

fromanirh:
failure due to github API rate limiting. Merging.

PR #104, opened 2019-11-07
Remove unnecessary commands from connect_to_rhel_console.exp file

ksimon1:
remove unnecessary commands from connect_to_rhel_console.exp file

### DIFF
--- a/automation/connect_to_rhel_console.exp
+++ b/automation/connect_to_rhel_console.exp
@@ -1,8 +1,9 @@
 #!/usr/bin/expect -f
 
+set timeout 300
+
 set kubeconfig [lindex $argv 0]
 set vm_name [lindex $argv 1]
+
 spawn ./virtctl --kubeconfig=$kubeconfig console $vm_name
-expect "Successfully connected to*\r"
-send -- "\r"
-expect "*login*\r"
+expect default { exit 1 } "*Welcome*\r"
--- a/automation/test-rhel.sh
+++ b/automation/test-rhel.sh
@@ -116,18 +116,10 @@ run_vm(){
     _oc describe vmi $vm_name
 
     set +e
-    current_time=0
-    while [  $(./connect_to_rhel_console.exp $kubeconfig $vm_name | grep login | wc -l) -lt 1 ] ; do 
-      echo "waiting for vm to be ready"
-
-      current_time=$((current_time + sample))
-      if [ $current_time -gt $timeout ]; then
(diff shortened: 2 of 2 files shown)


fromanirh:
Looks good, but please explain why the commands are unnecessary, and, if possible, since when they are unnecessary.

MarSik:
Yep, what @fromanirh said. The patch is good, just improve the commit message to explain what you did there (removed duplicated timeout loop, fixed too short default timeout of expect)...

ksimon1:
@MarSik, @fromanirh done

fromanirh:
Thanks! Almost there. Why 300 seconds? Is just a "big enough" value? In general a good commit message explains "the why" and provides context and backstory for the change. The "how" is usually self-evident from the code.
Everything else looks ok.

ksimon1:
@fromanirh it is default timeout defined here: https://github.com/kubevirt/common-templates/blob/master/automation/test-rhel.sh#L55

fromanirh:
> @fromanirh it is default timeout defined here: https://github.com/kubevirt/common-templates/blob/master/automation/test-rhel.sh#L55

Sure, I'm fine with the value. It's just that it deserved a mention, exactly like this comment, but in the commit message!

MarSik:
It is also the value we had before in the shell script.

MarSik:
@ksimon1 Also, please make the commit message lines shorter.

automation/connect_to_rhel_console.exp
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/expect -f
 
+set timeout 300
+timeout {exit 1 }
```
MarSik: I do not think it belongs here

automation/connect_to_rhel_console.exp
```diff
@@ -1,8 +1,10 @@
...
+
 set kubeconfig [lindex $argv 0]
 set vm_name [lindex $argv 1]
+
 spawn ./virtctl --kubeconfig=$kubeconfig console $vm_name
-expect "Successfully connected to*\r"
-send -- "\r"
-expect "*login*\r"
+expect "*Welcome*\r"
```
MarSik: Reading the man page suggests:
expect timeout { exit 1 } "*Welcome*\r"

MarSik: Or using the default that should catch both timeout and EOF

`expect default { exit 1 } "Welcome\r"`

PR #108, opened 2019-11-11
Remove template release tag from template name

MarSik:
The release tag is kept it the labels for template itself
and in the VM object the template describes. There is
no change there.

Adding the version tag to the template name was a mistake
that will break upgrades. The name should never change.

When an incompatible change is needed, a copy of a template
should be created, given new name and edited as needed.
The original must stay backwards compatible forever!


MarSik:
Most of the tests passed already and this was just a name change.

PR #109, opened 2019-11-11
Add fixed version suffix -v0.7.0 to all templates

MarSik:
This is a slight update to the recent patch that removed release
specific suffixes from template names. That fix forgot to take
into account the existing deployments that were done using SSP
operator and HCO.

The last version released via those channels was -v0.7.0 and
so to make upgrades from that release possible, I am readding
the suffix statically.

This can be interpreted as a version in which the template was
first officially released.

v0.7.0 is also the first version that appeared in a product
that supports upgrades. All versions prior that were always
static and had no upgrade guarantees.

Signed-off-by: Martin Sivak <msivak@redhat.com>

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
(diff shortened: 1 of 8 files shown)


fromanirh:
So looks like we are stuck with '-v0.7.0' in the names forever. Let's see if we can find a way to fix this in the future.

PR #114, opened 2019-12-04
remove unnecessary trailing comma from validation

ksimon1:
Linux templates have unnecessary trailing comma in validation
which causes error. This PR removes it.
//cc @fromanirh, @petrkotas 


fromanirh:
LGTM, filed https://github.com/kubevirt/common-templates/issues/115 to prevents issue like this in the future

PR #118, opened 2019-12-10
fix image downloading in prow tests

ksimon1:
This patch should fix the way, how images are downloaded and stored
in prow jobs. Previously, all images were stored in the same folder
under same name. It caused issues. This patch stores each image to its own folder. With this approach each job will have its own folder with image and
it prevents replacing image across jobs.


ksimon1:
//cc @petrkotas

Issue #115, opened 2019-12-04
tests: add a test to check all the validations are valid JSON


fromanirh:
the templates should ensure their validation annotation is syntactically valid, hence it is valid JSON.
Something like this could be enough (python like pseudocode):

```python
for tmpl in all(templates):
     validations = tmpl.annotations[TEMPLATE_VALIDATION]
    if not json.is_valid(validations):
       fail()
succeed()
```

ksimon1:
I was thinking about this as well. Maybe i can insert this check to e2e tests which we have now

PR #119, opened 2019-12-10
add json validation for validation in common-templates

ksimon1:
this validation checks if validation is in correct json format
Fixes https://github.com/kubevirt/common-templates/issues/115
/cc @petrkotas, @fromanirh 

### DIFF
--- a/travis_ci/check-validations.py
+++ b/travis_ci/check-validations.py
@@ -1,0 +1,45 @@
+#! python
+
+import logging
+import os
+import os.path
+import yaml
+import sys
+import json
+
+def checkValidations(path):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    for template in templates:
+        with open(path + "/" + template, 'r') as stream:
+            try:
+                template = yaml.safe_load(stream)
+
+                if template == None: 
+                    logging.info("Empty template file: %s", template)
+                    continue
+
+                logging.info("Checking " + template["metadata"]["name"])
+
+                try:
+                    json.loads(template["metadata"]["annotations"]["validations"])
+                except Exception as e:
+                    logging.info("Validation is not json")
+                    raise e
+
+            except yaml.YAMLError as exc:
+                raise exc
+
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


travis_ci/check-validations.py
```diff
@@ -0,0 +1,47 @@
+#! python
+
+import logging
+import os
```
fromanirh:
```python
import os.path
```
code as you wrote works, but in python we prefer to import the path subpackage explicitely

travis_ci/check-validations.py
```diff
@@ -0,0 +1,47 @@
...
+import yaml
+import sys
+import json
+
+def checkValidations(path):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    for template in templates:
+        with open(path + "/" + template, 'r') as stream:
```
fromanirh:
```python
with open(os.path.join(path, template)) as stream
```
'rt' mode is implicit

travis_ci/check-validations.py
```diff
@@ -0,0 +1,47 @@
...
+            try:
+                template = yaml.safe_load(stream)
+
+                if template == None: 
+                    logging.info("Empty template file")
```
fromanirh: please log the template name:
```python
logging.info("Empty or invalid template file: %s", template)
```

travis_ci/check-validations.py
```diff
@@ -0,0 +1,47 @@
...
+                    continue
+
+                metadata = template.get("metadata")
+                annotations = metadata.get("annotations")
+
+                logging.info("Checking " + metadata.get("name"))
+
+                try:
+                    json.loads(annotations.get("validations"))
```
fromanirh: no need to be so cautious: if this hierarchy is not present, we have a problem already:
```python
try:
    json.loads(template["metadata"]["annotations"]["validations"])
```

travis_ci/check-validations.py
fromanirh: OTOH explicit is better than implicit so feel free to keep the 'rt' mode

PR #120, opened 2020-01-13
Update Templates Memory Requests to Power-2-Based Values

orenc1:
There is a discrepancy between common-templates RAM requests for VMs vs. the required RAM which is being validated against osinfo-db.
This PR rectifies the templates memory requests to use the same units as the osinfo-db.

Reference:
https://bugzilla.redhat.com/show_bug.cgi?id=1781815

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -17,18 +17,18 @@
       src: rhel7.tpl.yaml
       dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}.yaml"
     with_items:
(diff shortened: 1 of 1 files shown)


MarSik:
Btw, I think someone should write a test that validates the sanity of conditions in CI. Just in case this happens again in the future.

ksimon1:
@MarSik there is already issue for condition tests https://github.com/kubevirt/common-templates/issues/115 (it is on my to-do list).
Failing travis tests are probably not related to this PR. But to be sure, i created fix for tests (https://github.com/kubevirt/common-templates/pull/121). After it gets merged, we can re-run tests

ksimon1:
@orenc1 can you please rebase this PR on latest master and force push changes?

PR #122, opened 2020-01-16
test win2019

ksimon1:
Do NOT merge.
This is test PR to run win 2019 test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,5 +92,3 @@ before_script:
 - bash -x ci/prepare-host virtctl $(bash -x ci/ci/extra/cat-kubevirt-release last)
 - bash -x ci/start-cluster $CPLATFORM
 - bash -x ci/deploy-kubevirt $CPLATFORM
-
-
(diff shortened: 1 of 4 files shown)


fabiand:
It seems as if the win2k12r2 template is used in the failing windows2019 test:
`cluster-up/kubectl.sh get vmi win2k12r2-server-medium -o json`

But shouldn't this use a windows2019 tempalte?
Also: I think the hyperv flags are used in the template. Can you try to drpo the flags for now?

ksimon1:
We use win2k12r2 template for all supported windows. Lets test it without hyperv flags.

fabiand:
How are we doing on this testing?

ksimon1:
@fabiand 
The win2019 is booting, but i had to remove whole clock section from template
So i am trying to identify which attribute is failing. And it means one test run for each attribute

ksimon1:
Closing in favour of: https://github.com/kubevirt/common-templates/pull/126

PR #125, opened 2020-01-29
rename win2k12r2 template to windows

ksimon1:
this change is needed, because win2k12r2 template is holding information
about 6 different versions of windows. General name windows is more suitable. win2k12r2 template is now deprecated and will be removed in next release.

//cc @petrkotas, @jelkosz


generate-templates.yaml
```diff
@@ -215,7 +215,19 @@
 
   - name: Generate Windows templates
     template:
-      src: win2k12r2.tpl.yaml
+      src: windows.tpl.yaml
+      dest: "{{ playbook_dir }}/dist/templates/windows-{{ item.workload }}-{{ item.flavor }}.yaml"
+    with_items:
+    - {flavor: medium, workload: server, memsize: "4Gi", cpus: 1, iothreads: False, tablet: False}
```
petrkotas: @ksimon1 are 4Gi enough to run the windows machine?

ksimon1: According to osinfo-db, win 2019 have recommended RAM size: https://gitlab.com/libosinfo/osinfo-db/blob/master/data/os/microsoft.com/win-2k19.xml.in#L26

petrkotas: Thanks, than I am OK with it.

PR #129, opened 2020-02-12
Removed tiny RHEL8 template 

omeryahud:
Removed tiny RHEL8 template as it does not comply with the minimal memory requirements

### DIFF
--- a/automation/test-rhel.sh
+++ b/automation/test-rhel.sh
@@ -64,6 +64,10 @@ if [[ $TARGET =~ rhel6.* ]]; then
   workloads=("desktop" "server")
 fi
 
+if [[ $TARGET =~ rhel8.* ]]; then
+  sizes=("small" "medium" "large")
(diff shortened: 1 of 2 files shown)


MarSik:
I see tiny specifies 1Gi of RAM and RHEL 8 requires 1073741824 (1Gi). Where do you see the issue? Also never remove templates that were released! There can be a VM that references them!

ksimon1:
Min memory size according to osinfo-db is 1610612736 (https://gitlab.com/libosinfo/osinfo-db/-/blob/master/data/os/redhat.com/rhel-8.0.xml.in#L128)
So if anyone tries to create vm from it, template-validator (if deployed) will throw an error.

MarSik:
It will not, because the value is retrieved from the metadata and you still have 1Gi there. You need to update the osinfo-db git submodule first to get the new values.

PR #140, opened 2020-03-31
Fix 'automation/test.sh' waiting for failed pod

akrejcir:
The image 'cockpit/kubernetes' is no longer in dockerhub,
which causes the pod 'registry-console' to fail to start.

With this patch, the pod is ignored when checking for readiness.

Signed-off-by: Andrej Krejcir <akrejcir@redhat.com>

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -210,12 +210,16 @@ fi
 timeout=300
 sample=30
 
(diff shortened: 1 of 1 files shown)


akrejcir:
@ksimon1 @omeryahud, can you take a look at this PR?

PR #141, opened 2020-04-02
change name.os.template.kubevirt.io annotation

ksimon1:
For UI purposes this PR changes name.os.template.kubevirt.io annotation to be in format: name.os.template.kubevirt.io/<os_short_name>: <os_nice_name> or higher VM

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -16,7 +16,7 @@
       /objects[0].spec.template.spec.networks
 
 {% for osl in oslabels %}
-    name.os.template.kubevirt.io/{{ osl }}: {{ lookup('osinfo', osl).name }}
+    name.os.template.kubevirt.io/{{ osl }}: {{ lookup('osinfo', oslabels[0]).name }} or higher VM
 {% endfor %}
 
     validations: |


MarSik:
But that might not be true. You cannot promise compatibility of future releases. 

For example: Fedora removed i386 support few releases ago. So in theory if you create a VM using Fedora 20 or higher and upgrade the system to Fedora 30, it might not boot anymore. (Of course this is a bit contrived, we only support 64 bit).

The correct way would be either a compatibility note in the UI (it is usual for newer releases to work as well) or a logic that properly computes the compatibility range.

Btw, we should probably check if the memory requirements are still the same for all releases. We assume that for Fedora templates iirc.

templates/_linux.yaml
```diff
@@ -16,7 +16,7 @@
       /objects[0].spec.template.spec.networks
 
 {% for osl in oslabels %}
-    name.os.template.kubevirt.io/{{ osl }}: {{ lookup('osinfo', osl).name }}
+    name.os.template.kubevirt.io/{{ osl }}: {{ lookup('osinfo', oslabels[0]).name }} or higher VM
```
MarSik, line 19: You should make sure the oslabels are sorted too. Just to make sure oslabels[0] is the oldest one.

ksimon1:
Shouldn't be better to just create new template fedora 32 and later and the old one set as e.g.  fedora 20 - fedora 31?

templates/_linux.yaml
ksimon1: oslabels are already sorted from previous task (https://github.com/kubevirt/common-templates/blob/master/generate-templates.yaml#L13)

PR #142, opened 2020-04-02
Fix rhel8 description

ksimon1:
rhel8 used rhel7 description. This PR adds rhel8 description


templates/rhel8.tpl.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
```
MarSik, line 4: Please update the name suffix to the actual version that will first contain this change.

PR #148, opened 2020-05-06
Fix BZ 1813198 - change recommended windows bus to sata

glekner:


glekner:
@ksimon1 can you review?

templates/win2k12r2-deprecated.tpl.yaml
```diff
@@ -45,8 +45,8 @@ metadata:
           "name": "windows-virtio-bus",
           "path": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
           "rule": "enum",
-          "message": "virto disk bus type has better performance, install virtio drivers in VM and change bus type",
-          "values": ["virtio"],
+          "message": "sata disk bus type has native support",
+          "values": ["sata"],
```
MarSik, line 49: This does not fix the bug from the title. The section you modified is just validation that prints out a warning to the user. It does not affect defaults.

The defaults are set elsewhere (line 119) and are already set to sata.

This change is also wrong. As far as I remember we want users to use virtio for disks, because that bus type has much better performance. The validation can be improved to not match cdroms of course, but it should still match disks I think.

glekner: Thanks for the response @MarSik 
There was a misunderstanding from my end, ultimately we want to force only CDs on Windows VMs to SATA only. we can for now force it in the UI but I believe there should be bus validations for CDs in common-templates. please correct me if i'm wrong.
If so, how do we proceed? should I open a bugzilla for it?

MarSik: You can add a second validation rule with the content you tried to change here, just update the jsonpath to `"path": "jsonpath::.spec.domain.devices.disks[*].cdrom.bus",`. Double check the cdrom format though as I have been away from kubevirt for many months.

glekner:
@MarSik can you take a look again? Thanks!

glekner:
@ksimon1 will this be included in 4.5? Thanks

ksimon1:
yes, this should fix be available in 4.5

PR #149, opened 2020-05-11
Add check for min memory consistency in generated templates

borod108:
Currently the minimum memory requirements set in the "validation" part of the generated template
might actually be smaller than the requirements of newest minor version of the os,
this script will through an error for such a template.

IMHO a better solution should be implemented in the future, but the obvious one - setting the
minimal memory requirement to the actual OS requirement will not do since we need to keep naming consistency,
i.e. "tiny" template means a specific memory requirement throughout the system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   include:
   - stage: Tests
-    name: Check syntax of all templates
+    name: Check syntax and memory requirements of all templates
     script: bash -x travis_ci/test_syntax.sh
 
   - name: Functional test of the Fedora template
--- a/travis_ci/test_syntax.sh
+++ b/travis_ci/test_syntax.sh
@@ -27,3 +27,4 @@ for template in $templates; do
 done
 
 python3 travis_ci/check-validations.py
+python3 travis_ci/validate-min-memory-consistency.py
--- a/travis_ci/validate-min-memory-consistency.py
+++ b/travis_ci/validate-min-memory-consistency.py
@@ -1,0 +1,66 @@
+#! python
+
+import logging
+import os
+import os.path
+import yaml
+import sys
+import json
+import importlib.util
+
+sys.path.insert(1, 'lookup_plugins/')
+import osinfo
+
+def minMemoryReqForOs(os_label):
+    latest_os_info = osinfo.LookupModule().run([os_label],[])[0]
+    return latest_os_info["minimum_resources.architecture=x86_64|all.ram"]
+
+def newestOsLabel(template):
+    labels = template["metadata"]["labels"]
+    osl_pref_str = "os.template.kubevirt.io"
+    os_labels = [label for label in labels if osl_pref_str in label]
+    return max(os_labels).split('/')[-1]
+
+def minMemoryReqInTemplate(template):
+    object = template["objects"][0]
+    min_str = object["spec"]["template"]["spec"]["domain"]["resources"]["requests"]["memory"]
+    min_gi_float = float(min_str.replace("Gi",""))
+    return int(min_gi_float * (1024**3))
+
+def checkMemoryReqs(path):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    for template_name in templates:
+        with open(os.path.join(path, template_name), 'r') as stream:
+            try:
+                template = yaml.safe_load(stream)
+
+                if template == None:
+                    logging.info("Empty template file: %s", template)
+                    continue
+
+                logging.info("Checking memory requirements consistency for: {}".format(template["metadata"]["name"]))
+
+                try:
+                    newest_os_label = newestOsLabel(template)
+                    actual_min_req = minMemoryReqForOs(newest_os_label)
+                    min_req_in_template = minMemoryReqInTemplate(template)
+                    if min_req_in_template < actual_min_req:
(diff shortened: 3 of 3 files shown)


borod108:
@omeryahud @ksimon1 First PR, please be gentle :)

omeryahud:
Hey Boris, looks good!

I noticed that the travis job failed with `ModuleNotFoundError: No module named 'ansible'` after running your script:
https://travis-ci.com/github/kubevirt/common-templates/jobs/331494510#L9702

You might have to install the python ansible module like so:
https://github.com/kubevirt/common-templates/blob/a3fe81256d754474ef1875914323d089d0f01462/.travis.yml#L88

Because the stack trace points to this line:
https://github.com/kubevirt/common-templates/blob/master/lookup_plugins/osinfo.py#L16

It would also be a good idea to integrate your solution to the e2e tests in `automation/test.sh`

ksimon1:
The code looks good to me. Please sign your commit with git commit -s and then force push your changes.

.travis.yml
```diff
@@ -86,6 +86,7 @@ addons:
     - intltool
 before_script:
 - pip install PyYAML
+- pip install ansible
```
cben, line 89: Hi, Boris asked me why it's failing, so commenting here.
I'm clueless about this project, may speak nonsense, feel free to ignore :)

My main suspicion is does this `pip` install for same python version as `python3` is later running?  Maybe it should be `pip3`?
I recommend writing `python3 -m pip install ...` to be more confident it matches.

Instead of installing specific packages here, consider keeping the list of dependencies with the code, e.g. travis_ci/requirements.txt and telling pip to use that.  Shouldn't matter, just more idiomatic.

(You could go further than that into using a venv for isolation, at which point it's worth adopting Poetry or Pipenv.  But then you won't benefit from system packages like `python-gobject` installed above, and Travis builds are disposable anyway, so I don't think you want this.)

borod108: than you very much Beni! actually adding the pip install ansible actually changed the error and now it is "gi" it is missing, I am just trying to figure out how not to make the tests run even longer by installing more packages, my feeling was that some of the tests all ready use osinfo.py and I am not adding anything that is not used there, so I was thinking I am not adding the test in the right place or something...

cben: Ah, I see the `No module named 'ansible'` was in an earlier build, seems `pip install ansible` did help :tada:.  I still recommend `python3 -m pip install` form.

cben:
New failure:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'gi'
```
is more of a pain to debug.   Do try running ansible with -vvv, but basically an seems ansible module can't import the `python-gobject` installed as an ubuntu package.  I assume this worked before?

Is it possible `pip install ansible` conflicts with the `ansible` ubuntu package?  Each using a different python and/or PYTHONPATH?
It may be bad idea to have both installed...
- Can you make travis_ci/*.py use the same python the system ansible uses?
- Or get ansible, and all packages it needs, only from pip.  Likely more work.

cben:
Short update: we found the problem is that Travis uses virtualenvs, which by default do NOT see system python libs installed with ubuntu `packages:`, such as `python-gobject`:
https://docs.travis-ci.com/user/languages/python/#travis-ci-uses-isolated-virtualenvs
So we believe `pip install ansible` added an `ansible-playbook` executable in the virtualenv's bin/ dir, which doesn't see any system libs.  (Previously, /usr/bin/ansible-playbook was running, which does see them.)
Obviously, having 2 ansible installs is confusing, want to settle on one.

The next hope was enable `system_site_packages` (see above doc) which lets the virtualenv see the system libs, but that only works with the system python version, which for `dist: xenial` (= ubuntu 16.04) is only 3.5 :-(

borod108:
@ksimon1 can you please take a look?
@cben thank you for your help on this PR!

borod108:
@ksimon1 can you please elaborate? Currently I compare the memory requirement with the newest minor version of the OS for every given template, going by the labels for that template.
Can the template be used to provision an OS that is not in the oslabels?
Or if there are templates for Centos6 and Centos7 and the memory requirement is wrong for centos8 I shoudl through the error for centos6 as well?

ksimon1:
Sorry I overlooked `return max(os_labels).split('/')[-1]` command. 
There is one issue: now it checks value in `metadata.annotations.validations` against value from osinfo-db. These two values will be always the same (see here: https://github.com/kubevirt/common-templates/blob/master/templates/_linux.yaml#L29).  It should check value in `objects[0].spec.template.spec.domain.resources.requests.memory` against value from osinfo-db

borod108:
@ksimon1 Thank you, you are right! I applied your fix, and now the tests will actually fail because: ERROR:root:Memory requirements for OS: ubuntu18.04 are not compatible with the requirements set in: ubuntu-desktop-tiny.yaml
So we actually caught one all ready :)

borod108:
@ksimon1 actually added a small improvement to see all the templates that have a problem at once, and there are 3 problematic ones.

ksimon1:
Thanks for the update. I already prepared PR which updates wrong values: https://github.com/kubevirt/common-templates/pull/153

borod108:
@ksimon1 awesome :) so you think this one can be merged after?

ksimon1:
Yes, please rebase this PR and after that we can merge it

PR #152, opened 2020-05-20
remove conroe cpu model from ubuntu template

ksimon1:
When creating a VM using the VM Wizard in the OpenShift Web Console,
if a user choose "Ubuntu 18.04 LTS", template forces Conroe CPU model.
This result in VM being unschedulable if this CPU type is not used
in the cluster.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1837670

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -205,7 +205,6 @@
     - {flavor: medium, workload: desktop, memsize: "4Gi", cpus: 1, iothreads: False, tablet: True}
     - {flavor: large, workload: desktop, memsize: "8Gi", cpus: 2, iothreads: False, tablet: True}
     vars:
-      cpumodel: Conroe
       os: ubuntu
       icon: ubuntu
       password: ubuntu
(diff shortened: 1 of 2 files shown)


omeryahud:
Don't forget to update the ubuntu template version, other than that looks good

MarSik:
@ksimon1 You should probably improve the commit message and explain the reason as the WHY is more important than the WHAT here. Iirc the cpu model was needed for Ubuntu to boot properly back when we introduced the templates.

ksimon1:
@MarSik  fixed

fabiand:
@ksimon1 @omeryahud there was a reason for the cpu model in the ubuntu template.

This ubuntu version had expectations on the underlying CPU model. It could be conroe or higher, thus if you are now launching this template on a cluster without the minimum model, then it will launch but the user will not see anything because the console stays black.

The solution is not to remove the model, but instead find the correct model which the image needs.

PR #154, opened 2020-05-20
Align processed VM's 'vm.kubevirt.io/template' label with template's name (include the tempalte's version)

omeryahud:
Template version is now defined in a variable in the template file in order to include the template version in `objects.metadata.labels["vm.kubevirt.io/template"]` and allow template validator to find the template.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1816518

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.7.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "CentOS 6.0+ VM"
     description: >-
--- a/templates/centos8.tpl.yaml
+++ b/templates/centos8.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.11.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.11.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "CentOS 8.0+ VM"
     description: >-
--- a/templates/fedora.tpl.yaml
+++ b/templates/fedora.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.7.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "Fedora 23+ VM"
     description: >-
--- a/templates/opensuse.tpl.yaml
+++ b/templates/opensuse.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.7.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "OpenSUSE Leap 15.0 VM"
     description: >-
--- a/templates/rhel6.tpl.yaml
+++ b/templates/rhel6.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.7.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "Red Hat Enterprise Linux 6.0+ VM"
     description: >-
--- a/templates/rhel7.tpl.yaml
+++ b/templates/rhel7.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.7.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "Red Hat Enterprise Linux 7.0+ VM"
     description: >-
--- a/templates/rhel8.tpl.yaml
+++ b/templates/rhel8.tpl.yaml
@@ -1,7 +1,8 @@
+{% set version =  "0.10.0" %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.10.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
   annotations:
     openshift.io/display-name: "Red Hat Enterprise Linux 8.0+ VM"
     description: >-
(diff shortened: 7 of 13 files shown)


templates/centos6.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: I think much better solution would be to set the version variable here (see line 14 for an example of how to set template variable). That will keep the version tied to the proper template file. And the include _linux.yaml will then propagete it to the common template.

Also, you should probably add a comment to the template explaining that the version must never change as it is part of the template identity. A change to version = change to VM compatibility.

generate-templates.yaml
```diff
@@ -35,6 +35,7 @@
       password: redhat
       oslabels: "{{ rhel8_labels }}"
       osinfoname: "{{ oslabels[0] }}"
+      version: 0.10.0
```
MarSik, line 38: Please do not do this here, it is not obvious when or how the version can be touched. I think it is much better to set the version in the template file itself. See my comment in the centos6 file.

templates/centos8.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.11.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/fedora.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/opensuse.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/rhel6.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/rhel7.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/rhel8.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.10.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/ubuntu.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v0.11.0
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/win2k12r2-deprecated.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: win2k12r2-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: win2k12r2-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/windows.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: windows-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: windows-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/windows10.tpl.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
-  name: windows10-{{ item.workload }}-{{ item.flavor }}-v0.7.0
+  name: windows10-{{ item.workload }}-{{ item.flavor }}-v{{ version }}
```
MarSik, line 4: Same as in centos6.

templates/centos6.tpl.yaml
omeryahud: That's a good idea, will do that, thanks!

omeryahud:
@MarSik could you take another look please?

templates/centos6.tpl.yaml
```diff
@@ -1,7 +1,10 @@
+# Change the version variable to the next released version of common-templates
+# if a non-backward compatible change was made to the template
```
MarSik, line 2: Well, there should be no incompatible changes to the template :) In case you need to do such change, create a new template with the same name, labels and new version.

omeryahud:
> Almost there :) Just fix the comment, the templates really should not be changed in any incompatible way. That would break the compatibility with already existing VMs linked to such templates.

From what I understood, old templates are not included in new releases.
Say the template `fedora-desktop-tiny-v0.1.0` is changed in a non-backward compatible way, and the version is incremented, only `fedora-desktop-tiny-v0.2.0` will be included in the next release, is that not right? or am I missing something?

MarSik:
Imagine you are restoring a backup. You will install a new cluster and then try to restore an old VM from somewhere. And you might need to have the template it was created with.

Now, you can install older version of templates first to get the old version, so it might be ok to just bump the version. But it would mean you have to take care of the ordering (old first) otherwise you would be overwriting some templates with their old revisions.

The original idea was to never remove anything, maybe it does not scale anymore and I can't decide that. Keeping stuff was simply easier at the time we started this project.

omeryahud:
> Imagine you are restoring a backup. You will install a new cluster and then try to restore an old VM from somewhere. And you might need to have the template it was created with.
> 
> Now, you can install older version of templates first to get the old version, so it might be ok to just bump the version. But it would mean you have to take care of the ordering (old first) otherwise you would be overwriting some templates with their old revisions.
> 
> The original idea was to never remove anything, maybe it does not scale anymore and I can't decide that. Keeping stuff was simply easier at the time we started this project.

I also think that keeping all the templates in the release is a better idea.
So whenever an update to a template is needed, I think the proper way to handle that update is to create a new template in `templates/*.tpl.yaml` (maybe copy the old `*.tlp.yaml` file and modify that, and increment it's version), and add an entry to generate it in `generate-templates.yaml` instead of just incrementing the template's version. That way every release includes all previously released templates along with their new versions.

Was this the flow that was planned originally?

MarSik:
Yes, that is exactly the flow I planned when I started this.

templates/centos6.tpl.yaml
omeryahud: I will remove this comment for now until I make sure the original flow is supported by the UI team and then I will prepare a new PR with comments and documentation

PR #156, opened 2020-06-09
Make template descriptions shorter

akrejcir:
The original descriptions were unnecessarily long.

Signed-off-by: Andrej Krejcir <akrejcir@redhat.com>

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -6,10 +6,8 @@ metadata:
   annotations:
     openshift.io/display-name: "CentOS 6.0+ VM"
     description: >-
-      This template can be used to create a VM suitable for
-      CentOS 6 and newer.
-      The template assumes that a PVC is available which is providing the
-      necessary CentOS disk image.
+      Template for CentOS 6 VM or newer.
+      A PVC with the CentOS disk image must be available.
     tags: "hidden,kubevirt,virtualmachine,linux,centos"
 
 {% set diskbus = diskbus |default("sata") %}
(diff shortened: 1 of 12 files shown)


templates/fedora.tpl.yaml
```diff
@@ -6,10 +6,8 @@ metadata:
   annotations:
     openshift.io/display-name: "Fedora 23+ VM"
     description: >-
-      This template can be used to create a VM suitable for
-      Fedora 23 and newer.
-      The template assumes that a PVC is available which is providing the
-      necessary Fedora disk image.
+      Template for Fedora 23 VM or newer.
```
ksimon1, line 9: Please change this to Fedora 31 VM or newer.
Please change `openshift.io/display-name` annotation as well

akrejcir: Done

templates/centos6.tpl.yaml
```diff
@@ -6,10 +6,8 @@ metadata:
   annotations:
     openshift.io/display-name: "CentOS 6.0+ VM"
     description: >-
-      This template can be used to create a VM suitable for
-      CentOS 6 and newer.
-      The template assumes that a PVC is available which is providing the
-      necessary CentOS disk image.
+      Template for CentOS 6 VM or newer.
+      A PVC with the CentOS disk image should be available.
```
MarSik, line 10: I believe "must be available" is the proper form here (the same in all the other templates).

akrejcir: Changed

akrejcir:
ping? Is this PR ok now?

PR #157, opened 2020-06-10
added cdrom disk to windows templates

ksimon1:
In windows templates is cdrom validation. Because there was no cdrom in disks,
validator returns error when user creates windows vm. This PR adds cdrom in vm disks.


MarSik:
Why force a cdrom when you can make the validation conditional? Use the `valid` part of the rule for that. See https://github.com/kubevirt/common-templates/blob/master/templates/VALIDATION.md#optional-keys

ksimon1:
> Why force a cdrom when you can make the validation conditional? Use the `valid` part of the rule for that. See https://github.com/kubevirt/common-templates/blob/master/templates/VALIDATION.md#optional-keys

Good point. I will update it.

templates/win2k12r2-deprecated.tpl.yaml
```diff
@@ -58,6 +58,7 @@ metadata:
         }, {
           "name": "windows-cd-bus",
           "path": "jsonpath::.spec.domain.devices.disks[*].cdrom.bus",
+          "valid": "jsonpath::.spec.domain.devices.disks[*].cdrom.bus",
```
MarSik, line 61: "jsonpath::.spec.domain.devices.disks[*].cdrom" is better, otherwise you will miss when somebody adds cdrom, but forgets to set the bus.

Also double check that [*] works for valid parameter.

ksimon1: fixed

PR #159, opened 2020-07-07
enable tablet device for windows server editions

ksimon1:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1827705


omeryahud:
@ksimon1 I think we can merge this

PR #161, opened 2020-07-17
Ignore disk validation if there are no disks

akrejcir:
The validation rules for disks should only be executed
if thre are some disks.

### DIFF
--- a/templates/win2k12r2-deprecated.tpl.yaml
+++ b/templates/win2k12r2-deprecated.tpl.yaml
@@ -43,6 +43,7 @@ metadata:
         }, {
           "name": "windows-virtio-bus",
           "path": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
+          "valid": "jsonpath::.spec.domain.devices.disks[*]",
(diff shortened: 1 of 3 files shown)


templates/win2k12r2-deprecated.tpl.yaml
```diff
@@ -43,13 +43,15 @@ metadata:
         }, {
           "name": "windows-virtio-bus",
           "path": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
+          "valid": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
```
MarSik, line 46:
```suggestion
          "valid": "jsonpath::.spec.domain.devices.disks[*].disk",
```

The validation should be applied even when the user forgets to set the bus.

akrejcir: Changed.

akrejcir:
The duplicated travis check looks like a github bug. Both checks point to the same build, which passed.

PR #162, opened 2020-07-20
[WIP] Change PVC to Datavolumes

ksimon1:
add datavolume labels
deprecated PVC templates

This PR deprecates old templates with PVC and replaces them with datavolumes labels

[better PR description and README update is still in progress.]

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -18,6 +18,10 @@
 {% for osl in oslabels %}
     name.os.template.kubevirt.io/{{ osl }}: {{ lookup('osinfo', oslabels[0]).name }} or higher
 {% endfor %}
+{% for osl in oslabels %}
+    pvc.template.kubevirt.io/{{ osl }}: "{{ osl }}"
+{% endfor %}
+    pvc.template.kubevirt.io/namespace: openshift-cnv-base-images
 
     validations: |
       [
@@ -38,7 +42,6 @@
     flavor.template.kubevirt.io/{{ item.flavor }}: "true"
     template.kubevirt.io/type: "base"
     template.kubevirt.io/version: "{{ lookup('env', 'VERSION') | default('devel', true) }}"
-
 objects:
 - apiVersion: kubevirt.io/v1alpha3
   kind: VirtualMachine
@@ -50,6 +53,22 @@ objects:
       vm.kubevirt.io/template.revision: "{{ lookup('env', 'REVISION') | default(1, true) }}"
       app: ${NAME}
   spec:
+    dataVolumeTemplates:
+      - apiVersion: cdi.kubevirt.io/v1beta1
+        kind: DataVolume
+        metadata:
+          name: ${NAME}
+        spec:
+          pvc:
+            accessModes:
+              - ReadWriteOnce
+            resources:
+              requests:
+                storage: 30Gi
+          source:
+            pvc:
+              name: ${PVCNAME}
+              namespace: ${PVCNAMESPACE}
     running: false
     template:
       metadata:
@@ -84,7 +103,7 @@ objects:
             disks:
             - disk:
                 bus: {{ diskbus | default("virtio") }}
-              name: rootdisk
+              name: ${NAME}
             - disk:
                 bus: {{ diskbus | default("virtio") }}
               name: cloudinitdisk
@@ -96,9 +115,9 @@ objects:
         - name: default
           pod: {}
         volumes:
-        - name: rootdisk
-          persistentVolumeClaim:
-            claimName: ${PVCNAME}
+        - dataVolume:
+            name: ${NAME}
+          name: ${NAME}
         - cloudInitNoCloud:
             userData: |-
               #cloud-config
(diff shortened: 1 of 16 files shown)


MarSik:
Is this backwards compatible? What will the UI do when the user tries to edit a VM created with an older version of a template that still uses PVC?

ksimon1:
> Is this backwards compatible? What will the UI do when the user tries to edit a VM created with an older version of a template that still uses PVC?

No it is not backward compatible. AFAIK UI is looking for templates by name, so it will load the older version of template.

MarSik:
> No it is not backward compatible. AFAIK UI is looking for templates by name, so it will load the older version of template.

Yep, I see you bumped the template version. I did not notice that when I asked the question.

templates/_linux.yaml
```diff
@@ -50,6 +53,24 @@ objects:
       vm.kubevirt.io/template.revision: "{{ lookup('env', 'REVISION') | default(1, true) }}"
       app: ${NAME}
   spec:
+    dataVolumeTemplates:
+      - apiVersion: cdi.kubevirt.io/v1alpha1
+        kind: DataVolume
+        metadata:
+          name: ${PVCNAME}
```
omeryahud, line 60: This looks problematic, if 2 or more VMs from the same template were to be generated, then a name collision would occur and the newer DV would fail to create or override the older one (depends on the implementation in kubevirt/kubevirt).
I think you should reuse `${NAME}` here, to guarantee that the DV name is generated and matches the VM's name

templates/_linux.yaml
```diff
@@ -50,6 +53,24 @@ objects:
...
+        spec:
+          pvc:
+            accessModes:
+              - ReadWriteOnce
+            resources:
+              requests:
+                storage: 1Gi
```
omeryahud, line 67: `1Gi` seems a bit conservative, we need to make sure that a template provides enough storage for the largest possible image that is used by it

templates/_linux.yaml
```diff
@@ -18,6 +18,10 @@
 {% for osl in oslabels %}
     name.os.template.kubevirt.io/{{ osl }}: {{ lookup('osinfo', oslabels[0]).name }} or higher
 {% endfor %}
+{% for osl in oslabels %}
+    datavolume.template.kubevirt.io/{{ osl }}: "{{ osl }}"
```
omeryahud, line 22: Maybe use the annotations `pvc.template.kubevirt.io/name` and `pvc.template.kubevirt.io/namespace` to make it explicit that the annotations point to a PVC? Some might think they point to a DV (I know we already agreed on `datavolume` before sorry about that :) )

templates/centos6.tpl.yaml
```diff
@@ -1,4 +1,4 @@
-{% set version =  "0.7.0" %}
```
omeryahud, line 1: I think we should add new entries to `generate-templates.yaml` instead of completely forgetting about the old templates.

Proposed flow:
1. Copy `centos6.tpl.yaml` to `centos6.tpl.v0.12.0.yaml`
1.1 Repeat 1 for all OSs
2. Modify `v0.12.0` templates to support the new DV flow
3. Add entries for those new templates in `generate-templates.yaml`
4. Reconcile both template versions for the following reasons:
- If a user of CNV 2.4 created a VM with a template, and for some reason, something happened to that template post 2.5 upgrade, the operator can reconcile that template.
- It gives us the flexibility of gracefully deprecating the templates in the future (by removing the
```
os.template.kubevirt.io/{{ osl }}: "true"
workload.template.kubevirt.io/{{ item.workload }}: "true"
flavor.template.kubevirt.io/{{ item.flavor }}: "true"
``` 
labels)

The downside of this is that we will be reconciling a large amount of templates over time, this might be solved if we decide to remove support for old templates for every major CNV version (3.0 for example)
What do you think?

ksimon1: I don't like the idea to have version in name of file. I would prefer to name it `centos6-dv.tpl.yaml / centos6-pvc.tpl.yaml`.

omeryahud: > I don't like the idea to have version in name of file. I would prefer to name it `centos6-dv.tpl.yaml / centos6-pvc.tpl.yaml`.

On the long run, this is problematic, if the dv templates are modified in the future, what would you call the new ones? dv-2?
Versioning the template files is the only consistent way I could think of that allows us to deprecate templates in the future, and because the DV templates are successors to the PVC templates, they should be a new version of them.
I think that you can also have directories for each version to make it more readable and maintainable, for example:
```
templates/
    | - v0.12.0/
            | - centos6.tpl.yaml
    | - v.0.11.0/
            | - centos6.tpl.yaml
```
and when a release comes, combine all templates from all versions into a single, what do you think?

omeryahud:
/hold 
Putting this on hold until we can get OCP CI to test it

ksimon1:
Closing in favour of https://github.com/kubevirt/common-templates/pull/274

PR #164, opened 2020-07-29
increase terminationGracePeriodSeconds to 30 seconds

ksimon1:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1861297

//cc @omeryahud, @rmohr 

**Release Note**:
```release-note
Increase terminationGracePeriodSeconds to allow guest VMs to safely shutdown before deleting the VMI pod
```

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -91,7 +91,7 @@ objects:
             interfaces:
             - masquerade: {}
               name: default
-        terminationGracePeriodSeconds: 0
+        terminationGracePeriodSeconds: 180
         networks:
         - name: default
           pod: {}
--- a/templates/windows10.tpl.yaml
+++ b/templates/windows10.tpl.yaml
@@ -1,4 +1,4 @@
-{% set version =  "0.7.0" %}
+{% set version =  "0.11.3" %}
 apiVersion: template.openshift.io/v1
 kind: Template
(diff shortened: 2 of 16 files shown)


templates/windows10.tpl.yaml
```diff
@@ -124,7 +124,7 @@ objects:
                 bus: usb
                 name: tablet
 {% endif %}
-        terminationGracePeriodSeconds: 0
+        terminationGracePeriodSeconds: 30
```
rmohr, line 127: I don't think that 30 seconds is a good default for windows. @ksimon1 what do you think? I would rather think into the direction of 1 hour or so.

templates/_linux.yaml
```diff
@@ -91,7 +91,7 @@ objects:
             interfaces:
             - masquerade: {}
               name: default
-        terminationGracePeriodSeconds: 0
+        terminationGracePeriodSeconds: 30
```
rmohr, line 94: Maybe give it two or three minutes?

templates/windows10.tpl.yaml
ksimon1: I think you are right. Shutting down windows can take much longer due to e.g. updates.

ksimon1:
@rmohr can you please review?

PR #166, opened 2020-08-02
Removed flavor/size labels from the deprecated windows template

omeryahud:
Fixed BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1856412

Removed the labels used to query for templates from the deprecated windows template.
If the labels are removed, the template will not show up when consumers look for a template to use for a new VM,
but, because it (the template) still exists, template validator can find it in order to use it for validation of old VMs.

Release note:

```release-note
Removed the flavor/size labels from deprecated templates so they will no longer show up when querying for templates
```


omeryahud:
/cc @ksimon1 @akrejcir @vatsalparekh 
Can you guys take a quick look?

vatsalparekh (approved):

PR #168, opened 2020-08-17
add new SRCPVCNAME and SRCPVCNAMESPACE params 

ksimon1:
add new SRC_PVC_NAME and SRC_PVC_NAMESPACE params to all templates.

These new params are added to unblock UI. They are not used in common templates
yet. In the future they will be used for datavolume template.

```release-note
add new SRC_PVC_NAME and SRC_PVC_NAMESPACE params to all templates.
```

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -114,4 +114,9 @@ parameters:
 - name: PVCNAME
   description: Name of the PVC with the disk image
   required: true
-
+- name: SRC_PVC_NAME
+  description: Name of the PVC to clone
+  value: '{{ oslabels | last }}'
+- name: SRC_PVC_NAMESPACE
+  description: Namespace of the source PVC
+  value: kubevirt-base-images
--- a/templates/windows.tpl.yaml
+++ b/templates/windows.tpl.yaml
@@ -147,3 +147,9 @@ parameters:
 - name: PVCNAME
   description: Name of the PVC with the disk image
   required: true
+- name: SRC_PVC_NAME
+  description: Name of the PVC to clone
+  value: win2k19
+- name: SRC_PVC_NAMESPACE
+  description: Namespace of the source PVC
+  value: kubevirt-base-images
--- a/templates/windows10.tpl.yaml
+++ b/templates/windows10.tpl.yaml
@@ -143,3 +143,9 @@ parameters:
 - name: PVCNAME
   description: Name of the PVC with the disk image
   required: true
+- name: SRC_PVC_NAME
+  description: Name of the PVC to clone
+  value: win10
+- name: SRC_PVC_NAMESPACE
+  description: Namespace of the source PVC
+  value: kubevirt-base-images


templates/_linux.yaml
```diff
@@ -114,4 +114,9 @@ parameters:
 - name: PVCNAME
   description: Name of the PVC with the disk image
   required: true
-
+- name: SRCPVCNAME
+  description: Name of the PVC to clone
+  value: '{{ oslabels | last }}'
```
fabiand, line 119: What values would occur here? Can you give a few examples?

ksimon1: e.g.
```
rhel8.2
rhel7.9
centos8
```

templates/_linux.yaml
```diff
@@ -114,4 +114,9 @@ parameters:
 - name: PVCNAME
   description: Name of the PVC with the disk image
   required: true
-
+- name: SRCPVCNAME
```
omeryahud, line 117: Can you replace `SRCPVCNAME` with `SRC_PVC_NAME` for better readability (same for namespace)?

templates/_linux.yaml
fabiand: Are all of these valid PVC names?
I suspect that anyething with a dot is not.

omeryahud: Names with dot are valid k8s names

templates/_linux.yaml
```diff
@@ -114,4 +114,9 @@ parameters:
 - name: PVCNAME
   description: Name of the PVC with the disk image
   required: true
-
+- name: SRC_PVC_NAME
+  description: Name of the PVC to clone
+  value: '{{ oslabels | last }}'
+- name: SRC_PVC_NAMESPACE
+  description: Namespace of the source PVC
+  value: openshift-kubevirt-base-images
```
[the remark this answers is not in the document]
ksimon1: Fine for me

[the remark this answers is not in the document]
ksimon1: fixed

omeryahud (approved):
/lgtm

omeryahud:
/cc @shwetaap 
Could you take a look at this Shweta? It would also help you figure out how you should update the default namespace downstream

shwetaap:
> /cc @shwetaap
> Could you take a look at this Shweta? It would also help you figure out how you should update the default namespace downstream
I see that the default namespace has been set to kubevirt-base-images as discussed. But where do we modify this to use the openshift namespace in downstream?

ksimon1:
> > I see that the default namespace has been set to kubevirt-base-images as discussed. But where do we modify this to use the openshift namespace in downstream?

This will be updated in d/s automation pipeline

shwetaap:
> > > I see that the default namespace has been set to kubevirt-base-images as discussed. But where do we modify this to use the openshift namespace in downstream?
> 
> This will be updated in d/s automation pipeline

ok got it. Thanks!

ksimon1:
/hold
currently we are trying to migrate common-templates to openshift-ci to be able to test this feature

PR #178, opened 2020-08-26
Update validation documentation

akrejcir:
**Release note**:
```release-note
Updated the documentation to include a non-intuitive behavior of "valid" field.
```

### DIFF
--- a/templates/VALIDATION.md
+++ b/templates/VALIDATION.md
@@ -146,10 +146,27 @@ kind: Template
 
 ### Optional Keys
 
-* `valid`: the rule must be *ignored* if the jsonpath given as value doesn't exist.
-*PLEASE NOTE* that even if values of this key are required to be JSONPaths, you still need to use the `jsonpath::` prefix
-as explained above.
 * `justWarning`: violating rule with justWarning field set will emit a warning only instead of failing the validation.
+* `valid`: the rule must be *ignored* if the jsonpath given as value doesn't exist.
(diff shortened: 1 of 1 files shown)


templates/VALIDATION.md
```diff
@@ -147,8 +147,10 @@ kind: Template
 ### Optional Keys
 
 * `valid`: the rule must be *ignored* if the jsonpath given as value doesn't exist.
-*PLEASE NOTE* that even if values of this key are required to be JSONPaths, you still need to use the `jsonpath::` prefix
-as explained above.
+  Some of the fields of the template have default values so they always exist, and setting this path to one of these fields
+  has no effect.
```
omeryahud, line 151: I know this can be tiresome, but can you please add some examples to such values? or at least values we already know of

omeryahud (changes requested):

templates/VALIDATION.md
akrejcir: All fields that are not defined as pointers, slices or maps in `VirtualMachineInstanceSpec` have default values. There are a lot of them. I could list a few, maybe it will be useful.

omeryahud: If you could list few of the important ones (like cpu) it would be great

akrejcir: Ok, I have added some fields that are probably important.

ksimon1 (approved):
/lgtm

PR #213, opened 2020-09-17
Add a parameter for randomized password

### LINKED ISSUE #187: Remove hardcoded passwords from templates
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
/kind bug
> /kind enhancement

**What happened**:

The templates configure vms to use password authentication, and include a hardcoded passwords. There are two separate security concerns here:

- Setting a default password means that people will start vms with that default configuration. If they are connected to a routeable network, they will rapidly be compromised.
- Passwords of any sort are vulnerable to brute force attempts

**What you expected to happen**:

I would expect the templates to have a template parameter for an SSH public key.

vatsalparekh:
Signed-off-by: Vatsal Parekh <vparekh@redhat.com>

Fixes #187 

**Release note**:

```release-note
Add a parameter for randomized password
```

### DIFF
--- a/templates/_linux.yaml
+++ b/templates/_linux.yaml
@@ -102,7 +102,7 @@ objects:
         - cloudInitNoCloud:
             userData: |-
               #cloud-config
-              password: {{ password }}
+              password: ${CLOUD_USER_PASSWORD}
               chpasswd: { expire: False }
           name: cloudinitdisk
 
@@ -120,3 +120,7 @@ parameters:
 - name: SRC_PVC_NAMESPACE
   description: Namespace of the source PVC
   value: kubevirt-os-images
+- description: Randomized password for cloud-user
+  from: '[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}'
+  generate: expression
+  name: CLOUD_USER_PASSWORD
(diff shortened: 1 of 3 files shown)


templates/_linux.yaml
```diff
@@ -102,7 +102,7 @@ objects:
         - cloudInitNoCloud:
             userData: |-
               #cloud-config
-              password: {{ password }}
+              password: ${PASSWORD}
```
ksimon1, line 105: As we discussed, please change this variable name to ${CLOUD_USER_PASSWORD}

templates/_linux.yaml
```diff
@@ -120,3 +120,7 @@ parameters:
 - name: SRC_PVC_NAMESPACE
   description: Namespace of the source PVC
   value: kubevirt-os-images
+- description: Randomized password
```
ksimon1, line 123: Change this to something like: `Randomized password for cloud user`

vatsalparekh: added

templates/_linux.yaml
vatsalparekh: changed

PR #233, opened 2020-09-29
Removed all OS labels from deprecated template

omeryahud:
Removed OS labels and annotations from the deprecated windows template in order to hide it from consumers

To prevent creation of new VMs from deprecated templates

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=1881458
https://bugzilla.redhat.com/show_bug.cgi?id=1856412

**Release note**:

```release-note
NONE
```

@rnetser @jelkosz @yaacov

### DIFF
--- a/templates/LABELS.md
+++ b/templates/LABELS.md
@@ -6,6 +6,12 @@ the workload a template is supposed to be used for.
 Some of them are not going to be used immediately and serve
 as examples.
 
+## Deprecated templates
+
+When a template is deprecated, the following annotation is applied to it: `template.kubevirt.io/deprecated: "true"`.
+
+The template itself is not removed for backward compatibility reasons.
+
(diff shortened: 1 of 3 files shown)


yaacov:
lgtm

MarSik:
Why not mark it as deprecated instead and fix the filtering on the UI side? I remember support status label idea we had ages ago.

MarSik:
Oh, I see you marked it as deprecated. Well if that is the way too you should probably document the new label/annotation in LABELS.md

omeryahud:
> Why not mark it as deprecated instead and fix the filtering on the UI side? I remember support status label idea we had ages ago.

The removal of the labels also makes it that `oc` users do not see this template when querying by labels

PR #235, opened 2020-09-30
Fix validation for disks

akrejcir:
The "valid" JSON path has to point to the "bus" field,
because "disk" field always exists, even if it is "nil".

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1883468

**Release note**:
```release-note
Fix validation rules for disk bus.
```


omeryahud (approved):
/lgtm
/approve

PR #256, opened 2020-10-14
Add a new template label to define a default variant for each OS

jean-edouard:
This label will allow the UI to offer a sensible default template for the selected operating system.
It effectively removes one VM creation step for users who don't have very specific requirements.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -17,18 +17,18 @@
       src: rhel8.tpl.yaml
       dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}.yaml"
     with_items:
-    - {flavor: tiny, workload: server, memsize: "1.5Gi", cpus: 1, iothreads: False, tablet: False}
-    - {flavor: tiny, workload: desktop, memsize: "1.5Gi", cpus: 1, iothreads: False, tablet: True}
-    - {flavor: tiny, workload: highperformance, memsize: "1.5Gi", cpus: 1, iothreads: True, tablet: False}
-    - {flavor: small, workload: server, memsize: "2Gi", cpus: 1, iothreads: False, tablet: False}
-    - {flavor: small, workload: desktop, memsize: "2Gi", cpus: 1, iothreads: False, tablet: True}
-    - {flavor: small, workload: highperformance, memsize: "2Gi", cpus: 1, iothreads: True, tablet: False}
-    - {flavor: medium, workload: server, memsize: "4Gi", cpus: 1, iothreads: False, tablet: False}
-    - {flavor: medium, workload: desktop, memsize: "4Gi", cpus: 1, iothreads: False, tablet: True}
-    - {flavor: medium, workload: highperformance, memsize: "4Gi", cpus: 1, iothreads: True, tablet: False}
-    - {flavor: large, workload: server, memsize: "8Gi", cpus: 2, iothreads: False, tablet: False}
-    - {flavor: large, workload: desktop, memsize: "8Gi", cpus: 2, iothreads: False, tablet: True}
-    - {flavor: large, workload: highperformance, memsize: "8Gi", cpus: 2, iothreads: True, tablet: False}
+    - {flavor: tiny, workload: server, memsize: "1.5Gi", cpus: 1, iothreads: False, tablet: False, default: False}
+    - {flavor: tiny, workload: desktop, memsize: "1.5Gi", cpus: 1, iothreads: False, tablet: True, default: False}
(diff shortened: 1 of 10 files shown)


omeryahud (changes requested):
I knew I forgot to submit the review :)

generate-templates.yaml
```diff
@@ -46,18 +46,18 @@
...
-    - {flavor: medium, workload: server, memsize: "4Gi", cpus: 1, iothreads: False, tablet: False}
-    - {flavor: medium, workload: desktop, memsize: "4Gi", cpus: 1, iothreads: False, tablet: True}
-    - {flavor: medium, workload: highperformance, memsize: "4Gi", cpus: 1, iothreads: True, tablet: False}
-    - {flavor: large, workload: server, memsize: "8Gi", cpus: 2, iothreads: False, tablet: False}
-    - {flavor: large, workload: desktop, memsize: "8Gi", cpus: 2, iothreads: False, tablet: True}
-    - {flavor: large, workload: highperformance, memsize: "8Gi", cpus: 2, iothreads: True, tablet: False}
+    - {flavor: tiny, workload: server, memsize: "1Gi", cpus: 1, iothreads: False, tablet: False, default: False}
+    - {flavor: tiny, workload: desktop, memsize: "1Gi", cpus: 1, iothreads: False, tablet: True, default: False}
+    - {flavor: tiny, workload: highperformance, memsize: "1Gi", cpus: 1, iothreads: True, tablet: False, default: False}
+    - {flavor: small, workload: server, memsize: "2Gi", cpus: 1, iothreads: False, tablet: False, default: False}
```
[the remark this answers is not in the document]
jean-edouard: Oops, good catch, will fix, thanks!

MarSik:
Uh, honestly, I do not understand the use case and the expected user flow here. The expected flow always was that you select the OS first and then look up the right template using the template labels.

MarSik:
It might make sense if you want to give the user a default sized windows VM. It seems the patch does that indeed. So maybe explain the expected flow in the commit message?

jean-edouard:
Added more info to the first commit and updated the PR description accordingly.

travis_ci/test_defaults.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+set -ex
+
+templates_dir='dist/templates'
+default_label='template.kubevirt.io/default-os-variant=true'
+
+make generate
+oc create -f $templates_dir
+
+for os in `(cd $templates_dir && ls -1 *.yaml) | cut -d '-' -f 1 | sort -u`; do
```
omeryahud, line 27: It is not enough to only iterate over the os portion of the template name, because every template covers multiple operating systems, for example, `rhel8` templates have these labels:
```
os.template.kubevirt.io/rhel8.0: "true"
os.template.kubevirt.io/rhel8.1: "true"
os.template.kubevirt.io/rhel8.2: "true"
os.template.kubevirt.io/rhel8.3: "true"
```
So the loop should iterate over all `os.template.kubevirt.io/*` within each template

omeryahud (changes requested):
Just a small issue with the test, otherwise looks good

travis_ci/test_defaults.sh
jean-edouard: Got it, new version pushed!

omeryahud (approved):
/lgtm

omeryahud:
@ksimon1 Can you give this a review as well please?

templates/deprecated-windows.tpl.yaml
```diff
@@ -62,6 +62,9 @@ metadata:
   labels:
     template.kubevirt.io/type: "base"
     template.kubevirt.io/version: "{{ lookup('env', 'VERSION') | default('devel', true) }}"
+{% if item.default %}
```
ksimon1, line 65: I would not add this label to deprecate templates

omeryahud:
@jean-edouard Please ignore all openshift-ci lanes for this PR, they are still WIP.
I noticed an issue with the fedora test on travis, I have posted a PR to hopefully resolve that, if all other travis tests pass, We won't block merging on it

templates/deprecated-windows.tpl.yaml
omeryahud: I have a PR open to remove those templates, shouldn't be a problem for 0.13.0

omeryahud: https://github.com/kubevirt/common-templates/pull/275

omeryahud:
All Travis tests passed (except for Fedora, which was expected), manually merging this

PR #274, opened 2020-10-21
OCP CI integration and dataVolumeTemplates

ksimon1:
Switch main testing platform to openshift CI.
Switch from PVC storage to dataVolumeTemplates for all templates.
    
This PR prepares common templates for switching to openshift CI. In builder folder
is located dockerfile which is used as a base image for openshift ci job. This dockerfile
contains all necessary sw which is needed to generate templates. This PR removes unnecessary
scripts (like automation/x-limit-ram-size.sh, create-minikube-pvc.sh, ...). These scripts were used only
on travis and currently are not used anywhere. kubevirtci and ci submodules are removed as well,
because templates will now use openshift 4.6. All tests (free and non free OSes) will now run only
on openshift ci. Travis tests are removed and only release job stays.
    
This PR removes pvc storage from all templates and replaces it with dataVolumeTemplates.
User now has to create its own PVC / datavolume.

PR which disables prow https://github.com/kubevirt/project-infra/pull/678

**Release note**:
```
Common templates are now using dataVolumeTemplates
```


ksimon1:
Retest will not work, because test script is converted for openshift ci and these tests still runs on prow.

omeryahud:
Let's merge the openshift/release PR first, then consider merging this when we see all OCP ci lanes passing.

automation/test.sh
```diff
@@ -170,143 +40,111 @@ for filename in dist/templates/*; do
...
+  name: cdi-role
+  namespace: cdi
+rules:
+- apiGroups: ["cdi.kubevirt.io"]
+  resources: ["datavolumes/source"]
+  verbs: ["*"]
+---
+EOF
+
+export SSP_VERSION=$(curl -s https://api.github.com/repos/kubevirt/kubevirt-ssp-operator/releases | \
```
omeryahud, line 116: I would prefer deploying template validator without the operator, it allows us to consume released versions before the reach the operator

omeryahud (changes requested):

automation/test.sh
ksimon1: I am preferring deployment with ssp operator because:
1) we don't have to deal with certificates (sometimes whole test failed due to certificate issue). 
2) Templates will be tested with validator which is shipped with HCO
3) validator deploy manifests in original repo are not prepared to be deployed on OCP 4 yet

ksimon1:
/hold
I have to create new test, which will run all unit tests on openshift ci, so we can get rid of travis

ksimon1:
/hold cancel
This PR https://github.com/openshift/release/pull/13118 has to be merged before this PR

automation/test-linux.sh
```diff
@@ -0,0 +1,182 @@
...
+  vm_name=$1
+  template_path="dist/templates/$vm_name.yaml"
+  local template_name=$( oc get -n ${namespace} -f ${template_path} -o=custom-columns=NAME:.metadata.name --no-headers -n kubevirt )
+  running=false
+
+  #If first try fails, it tries 2 more time to run it, before it fails whole test
+  for i in `seq 1 3`; do
+    error=false
+    oc process -n $namespace -o json $template_name NAME=$vm_name SRC_PVC_NAME=$TARGET-datavolume-original SRC_PVC_NAMESPACE=kubevirt | \
+    jq 'del(.items[0].spec.dataVolumeTemplates[0].spec.pvc.accessModes) |
```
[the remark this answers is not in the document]
ksimon1: this is test, user should not modify this

templates/_linux.yaml
```diff
@@ -99,9 +115,9 @@ objects:
         - name: default
           pod: {}
         volumes:
-        - name: rootdisk
-          persistentVolumeClaim:
-            claimName: ${PVCNAME}
+        - dataVolume:
```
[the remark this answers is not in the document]
ksimon1: no, only via datavolume

templates/_linux.yaml
```diff
@@ -53,6 +53,22 @@ objects:
...
+      kind: DataVolume
+      metadata:
+        name: ${NAME}
+      spec:
+        pvc:
+          accessModes:
+            - ReadWriteMany
+          resources:
+            requests:
+              storage: 15Gi
```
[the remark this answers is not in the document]
ksimon1: this size was chosen during the tests, because e.g. CentOS required about 15Gi.

automation/test-linux.sh
```diff
@@ -0,0 +1,182 @@
...
+kind: DataVolume
+metadata:
+  name: ${TARGET}-datavolume-original
+spec:
+  source:
+    registry:
+      url: "${image_url}"
+      ${secret_ref}
+  pvc:
+    accessModes:
```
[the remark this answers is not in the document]
fabiand: In the context of these tests they can stay undefined.
We are still working with Adam in order to finalize the whoel story how applying SC and VM will be managed.

automation/test-linux.sh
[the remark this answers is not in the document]
fabiand: Today the testing env does not provide RWX volumes, thus we have to do this small shortcut in odrder to run the tests.
It's good if we can provide RWX volumes in future, but right now it is okay to do this small change.

The templates themselves are build with the assumption of shared storage, thus the templates need to maintain the need for RWX.
But for this testcase the mode is also irrelevant (we are simply testing if guests come up or not).

templates/_linux.yaml
```diff
@@ -53,6 +53,22 @@ objects:
       vm.kubevirt.io/template.revision: "{{ lookup('env', 'REVISION') | default(1, true) }}"
       app: ${NAME}
   spec:
+    dataVolumeTemplates:
+    - apiVersion: cdi.kubevirt.io/v1beta1
+      kind: DataVolume
+      metadata:
+        name: ${NAME}
+      spec:
+        pvc:
```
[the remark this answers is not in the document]
fabiand: Yes, but this is done in conjucntion with Adam et al.

templates/_linux.yaml
```diff
@@ -99,9 +115,9 @@ objects:
         - name: default
           pod: {}
         volumes:
-        - name: rootdisk
-          persistentVolumeClaim:
-            claimName: ${PVCNAME}
+        - dataVolume:
+            name: ${NAME}
```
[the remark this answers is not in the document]
fabiand: To me it's ok to call the DV the same as the VM, it might actually help, because the same DV name maps to the same VM name.

templates/_linux.yaml
```diff
@@ -53,6 +53,22 @@ objects:
...
+      metadata:
+        name: ${NAME}
+      spec:
+        pvc:
+          accessModes:
+            - ReadWriteMany
+          resources:
+            requests:
+              storage: 15Gi
+        source:
```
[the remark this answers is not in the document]
fabiand: It's intentional that we change the flow to be focused on cloning via DVs, this is safer.

Users can still choose PVCs - very easily in the UI, more cumbersome on the CLI.
If it's a common case for them, then users are asked to clone the template and adjust it to their likings.

templates/_linux.yaml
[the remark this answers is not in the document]
ksimon1: After discussion with Fabian, I updated linux templates to have 30Gi of storage

automation/test.sh
omeryahud: For the next release, we have changes for both validator and templates, if we use the operator here, it means the operator has to consume the new TV before it was tested with templates, which is not a good idea.
IMO the proper way would be to test templates with validator first, and then let the operator consume them

automation/test.sh
```diff
@@ -170,143 +40,107 @@ for filename in dist/templates/*; do
...
+kind: KubevirtTemplateValidator
+metadata:
+  name: kubevirt-template-validator
+  namespace: kubevirt
+spec:
+  templateValidatorReplicas: 2
+---
+EOF
+
+while [ $(oc get pods -n $namespace  | grep validator | wc -l) -eq 0 ] ; do
```
omeryahud, line 129: You should wait for the `Available` conditin on the validator deployment if you are waiting for it to be ready to accept requests

automation/unit-tests.sh
```diff
@@ -0,0 +1,40 @@
...
+if [ $? -eq 1 ];then
+  echo "Validation of duplicate templates failed "
+  exit 1
+fi
+
+./automation/test_defaults.sh 
+if [ $? -eq 1 ];then
+  echo "Validation of default label failed "
+  exit 1
+fi
```
omeryahud, line 40: Missing newline

omeryahud (changes requested):

automation/test-linux.sh
```diff
@@ -0,0 +1,182 @@
...
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 30Gi
+EOF
+
+timeout=600
+sample=10
+current_time=0
+while [ $(oc get pods -n $namespace | grep "${TARGET}-datavolume-original.*Running" | wc -l ) -eq 0 ] ; do
```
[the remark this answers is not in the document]
akrejcir: Or at least can you combine these two loops into one, with condition '< 1', instead of '== 0' and then '== 1' ?

automation/test-linux.sh
```diff
@@ -0,0 +1,182 @@
...
+EOF
+
+timeout=600
+sample=10
+current_time=0
+while [ $(oc get pods -n $namespace | grep "${TARGET}-datavolume-original.*Running" | wc -l ) -eq 0 ] ; do 
+  oc get pods -n $namespace
+  current_time=$((current_time + sample))
+  if [ $current_time -gt $timeout ]; then
+    error=true
```
akrejcir, line 43: This `error` variable is set here, but it is ignored and the test continues as if no error happened.

automation/test-linux.sh
```diff
@@ -0,0 +1,182 @@
...
+    jq 'del(.items[0].spec.dataVolumeTemplates[0].spec.pvc.accessModes) |
+    .items[0].spec.dataVolumeTemplates[0].spec.pvc+= {"accessModes": ["ReadWriteOnce"]} | 
+    .items[0].metadata.labels["vm.kubevirt.io/template.namespace"]="kubevirt"' | \
+    oc apply -n $namespace -f -
+
+    # start vm
+    ./virtctl start $vm_name -n $namespace
+    # wait until import pod is running
+    current_time=0
+    while [ $(oc get pods -n $namespace | grep "cdi-upload-$vm_name.*Running" | wc -l ) -eq 0 ] ; do
```
akrejcir, line 116: There can be a race condition here, if the upload pod finishes quickly, before this loop can detect it.

automation/test-linux.sh
```diff
@@ -0,0 +1,182 @@
...
+      if [ $current_time -gt $timeout ]; then
+        error=true
+        break
+      fi
+      sleep $sample;
+    done
+
+    # wait until import is finished
+    current_time=0
+    while [ $(oc get pods -n $namespace | grep "cdi-upload-$vm_name.*Running" | wc -l ) -gt 0 ] ; do
```
akrejcir, line 128: These loops don't need to run, if the `error` variable is true. It would just add more time to the test for no reason.

automation/test-windows.sh
```diff
@@ -2,78 +2,76 @@
...
 restartPolicy: Always
 ---
 EOF
 
 timeout=1000
-sample=30
+sample=10
+current_time=0
+#check if cdi import pod is running
+while [ $(oc get pods -n $namespace | grep "${TARGET}-datavolume-original.*Running" | wc -l ) -eq 0 ] ; do
```
akrejcir, line 48: See all my comments for the `test-linux.sh` file. They apply here too.

akrejcir (changes requested):
It seems these files are not called from anywhere:
- `automation/test_syntax.sh`
- `travis_ci/test.sh`

README.md
```diff
@@ -26,17 +26,14 @@ Is it necessary to install the following components to be able to run the Ansibl
...
-provided (i.e.`PVCNAME`).
 
 ```bash
 # Clone the repository
 $ git clone https://github.com/kubevirt/common-templates
 $ cd common-templates
 
 # Pull all submodules
-$ git submodule init
-$ git submodule update
```
omeryahud, line 39: I think `git submodule update` is still required, not sure though

README.md
```diff
@@ -78,13 +71,17 @@ The table below lists the guest operating systems that are covered by the templa
 
 | Guest OS | Meta-template |
 |---|---|
-| Microsoft Windows Server 2012 R2 | [windows](templates/windows.tpl.yaml) |
+| Microsoft Windows Server 2012 R2 | [windows](templates/windows2k12.tpl.yaml) |
+| Microsoft Windows Server 2016 | [windows](templates/windows2k16.tpl.yaml) |
+| Microsoft Windows Server 2019 | [windows](templates/windows2k19.tpl.yaml) |
 | Microsoft Windows 10 | [windows](templates/windows10.tpl.yaml) |
-| Fedora 32 | [fedora](templates/fedora.tpl.yaml) |
+| Fedora 39 | [fedora](templates/fedora.tpl.yaml) |
```
omeryahud, line 78: I think it is a bit early to support Fedora 39

templates/README.md
```diff
@@ -138,9 +139,13 @@ objects:
...
       spec:
         domain:
+          cpu:
+            sockets: 2
+            cores: 1
+            threads: 1
           resources:
             initial:
-              memory: ${MEMORY_SIZE}Mi
+              memory: 256Mi
```
omeryahud, line 148: Maybe `256Mi` is not a good example for a win10 VM

templates/README.md
```diff
@@ -234,9 +239,13 @@ spec:
...
     spec:
       domain:
+        cpu:
+          sockets: 2
+          cores: 1
+          threads: 1
         resources:
           initial:
-            memory: 8Mi
+            memory: 256Mi
```
omeryahud, line 248: Should also be updated according to the above value

omeryahud (changes requested):
Minor doc issues, other than that lgtm

README.md
ksimon1: It is good to be prepared :D

automation/test-linux.sh
```diff
@@ -0,0 +1,126 @@
...
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 30Gi
+EOF
+
+timeout=600
+sample=10
+
+oc wait --for=condition=Ready --timeout=${timeout}s dv/${TARGET}-datavolume-original -n $namespace
```
akrejcir, line 39: Should the tests fail, if these `oc wait` commands time out?

ksimon1: in this case yes, because otherwise it will not import the base image

omeryahud:
/lgtm cancel
I see lots of `Unable to use a TTY - input is not a terminal or the right kind of file` in the windows tests, do you know the reason for it?

ksimon1:
in openshift ci you can't enable TTY interactive mode, that is why it shows this warning. It doesn't mean the tests are wrong. I updated the command and it should not show this warning anymore.

automation/test.sh
```diff
@@ -170,143 +40,98 @@ for filename in dist/templates/*; do
...
+            jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
+oc apply -f https://github.com/kubevirt/containerized-data-importer/releases/download/$CDI_VERSION/cdi-operator.yaml
+oc apply -f https://github.com/kubevirt/containerized-data-importer/releases/download/$CDI_VERSION/cdi-cr.yaml
 
-  if [ $webhookUpdateRetries == 0 ];
-  then
-    echo Retry count for template validator ca bundle injection reached
-    exit 1
-  fi
+oc wait --for=condition=Available --timeout=${timeout}s deployment/cdi-operator -n cdi
```
omeryahud, line 98: I think you can wait for the CDI CR instead of independent deployments

automation/test-windows.sh
```diff
@@ -103,85 +67,49 @@ delete_vm(){
...
+    jq 'del(.items[0].spec.dataVolumeTemplates[0].spec.pvc.accessModes) |
+    .items[0].spec.dataVolumeTemplates[0].spec.pvc+= {"accessModes": ["ReadWriteOnce"]} | 
+    .items[0].metadata.labels["vm.kubevirt.io/template.namespace"]="kubevirt"' | \
+    oc apply -f -
+    
     # start vm
-    ./virtctl --kubeconfig=$kubeconfig start $vm_name
+    ./virtctl start $vm_name -n $namespace
+    sleep 10
+    oc wait --for=condition=Ready --timeout=${timeout}s dv/$vm_name -n $namespace
```
omeryahud, line 96: I think I saw one of the CI lanes fail here because the DV was not created yet. 
You can drop this `oc wait` because it is guaranteed that a VMI will not start until the DV that was assigned to it is ready, so `oc wait`ing for the `vm` condition is enough

omeryahud (commented):
I found a possibly flakey command

omeryahud (commented):
/lgtm

omeryahud (approved):

PR #278, opened 2020-10-22
Improve tempalte maintainability

omeryahud:
This PR changes the upgradability of templates.

Until now, we had to generate new templates each time a non-backward compatible change was made to a template. This method floods the user's cluster with older templates over time that are not reconciled.

In order to improve the situation, this PR includes the following changes:
- Removal of deprecated templates: A user shouldn't have deprecated templates deployed on their cluster, so they were removed from the bundle. If an older VM still needs the template, it will exist in the cluster from a previous deployment, we will not actively remove it.
- Embedding the `_linux.yaml` template into all OS variant templates: In order to have the ability to provide templates that are fine-tuned to each OS needs, the `_linux.yaml` template was integrated into them to allow changes that will not affect other OS's.
- Moving the `validation` annotation to the VM resource from the Template: In order to allow templates to be updated in the future, we needed to break the dependency of VMs on their parent template. This mechanism allows us to easily update templates because each VM holds it's own validation rules and can be verified in-place (a supporting PR is open for `template-validator`: https://github.com/kubevirt/kubevirt-template-validator/pull/53).

**Release note**:

```release-note
Validation annotation moved from the Template to the processed VM
```


jean-edouard:
When working on templates, I really appreciated the fact that the values that are common to all Linux templates were factored into just one file. It made it easy to add stuff to all Linuxes, while highlighting what specific OSes do differently.
What is the motivation for "un-factoring" everything?
I'm not strongly against it, just wondering.
The rest of the PR LGTM.

omeryahud:
> When working on templates, I really appreciated the fact that the values that are common to all Linux templates were factored into just one file. It made it easy to add stuff to all Linuxes, while highlighting what specific OSes do differently.
> What is the motivation for "un-factoring" everything?
> I'm not strongly against it, just wondering.
> The rest of the PR LGTM.

The motivation is to provide a cleaner way to modify a template without affecting others, without it, an optimization for a specific OS requires another field added to the `generate-templates.yaml` templating items, and an exception in the `_linux.yaml` template that looks like this:
```
{{% if item.X %}}
    [whatever field that's relevant for a specific os]
{{% endif %}}
```
It turnes the `_linux.yaml` template to a very unreadable one in the long run.

jean-edouard (approved):

omeryahud:
Rebasing is quite complex, will push the updates soon

PR #282, opened 2020-11-02
Improve template maintainability

omeryahud:
This PR presents the following changes:

- **Removal of deprecated templates**:
	- Templates that are no longer in use should not be deployed on the
	client's cluster, so they were removed (if they are needed for old
	VMs, they exist on a cluster from a previous deployment.

- **_linux.yaml Embedded into variant templates**:
	- In order to support fine-tuning of specific linux OSs without
	affecting others, the _linux.yaml template was embedded into
	all linux variants templates

- **vm.kubevirt.io/validation annotation was introduced**:
	- In order to provide an easy upgrade for templates, the 'validation'
	label was moved to the processed VM from the parent template.
	This change breaks the dependency of the VM in the template
	during validation, and allows us the freedom to upgrade
	templates

**Release note**:

```release-note
Validation annotation was moved to the VM resource from the parent Template
```


MarSik:
The second and third changes go against the initial design decisions. Do you have a design doc explaining the benefits?

For the embedding part: Is there any fine tuning needed for linux templates at this moment? If not, is this a premature optimization that will cost you more time when updating something global to all linux templates?

About the validation. This will complicate bug fixes and VM edits. The current way expects that a template exists for any VMs that are using it and you have to maintain it, but your new way has issues in how you propagate changes:

- What happens when you find a bug in the template?
- How will the new and fixed validation propagate to already created VMs?
- Which version of the validation will win when the user tries to edit a VM and the template validations do not agree with what is stored in the VM?

omeryahud:
Hi Martin, thanks for the review!

> The second and third changes go against the initial design decisions. Do you have a design doc explaining the benefits?

No, I don't have a design document, I will try to explain myself with answers to your questions below

> For the embedding part: Is there any fine tuning needed for linux templates at this moment? If not, is this a premature optimization that will cost you more time when updating something global to all linux templates?

Yes, this is a premature optimization. I am not worried about global modifications to linux templates, because from my experience, both cases are pretty rare and having a centralized template for each operating system makes template editing easier IMO
 
> About the validation. This will complicate bug fixes and VM edits. The current way expects that a template exists for any VMs that are using it and you have to maintain it, but your new way has issues in how you propagate changes:
> 
> * What happens when you find a bug in the template?

We would update the template in the next release of common-templates, and that fix would only affect newly created VMs.

> * How will the new and fixed validation propagate to already created VMs?

IMO we should never retroactively affect a VM's validation rules or spec fields post creation.
Imagine a user creates a VM using a buggy template X with a buggy validation rule, with the current mechanism, two things could have happened:
* We fix the validation rule and create a new template (bump template version) - VM is not affected
* We fix the validation directly on the existing template - VM is affected

I think the latter is something we should never do, because the VM could have been edited by the user by the time the fix was shipped and we might retroactively break existing workloads

With the mechanism I have proposed here, we will fix the faulty validation rule, and that fix would affect newly created VMs only (because VMs are shipped with validation rules embedded)

> * Which version of the validation will win when the user tries to edit a VM and the template validations do not agree with what is stored in the VM?

This mechanism suggests to replace the template validation mechanism, so if a VM has validations annotation embedded, those validation rules would be used, otherwise, the template validation rules would be used (only relevant for VMs created with the current and previous releases of common-templates - backward compatibility)

I am unsure about safely upgrading VMs from template version X to template version X+1 (in case of a bugfix or a feature update), but I do not think this should be an automatic action, I think the user should actively confirm the upgrade and understand the circumstances, instead of us potentially breaking their workloads.

MarSik:
> Yes, this is a premature optimization. I am not worried about global modifications to linux templates, because from my experience, both cases are pretty rare and having a centralized template for each operating system makes template editing easier IMO

Well, it might be ok nowadays. We actually started with separate files, but then found out that the differences were rather small. After that there were global changes to all linux templates in the early phases. You could call that the dust settling down.

Myself I prefer the current approach that uses includes. I just like the DRY (do not repeat yourself) aspect of it.
 
> > * How will the new and fixed validation propagate to already created VMs?
> 
> IMO we should never retroactively affect a VM's validation rules or spec fields post creation.

I agree with the spec part. I think I even said so in templates/README.md#template-edits

I would argue that metadata can be changed. Like in the RPM world's NVR - version and sources are fixed, the metadata can change and the R (release) number then increases.

We had at least one case where the validations were wrong (due to an old osinfo-db) and fixing the validation actually made sense (iirc it was about RHEL 8 memory requirements).

> * We fix the validation rule and create a new template (bump template version) - VM is not affected
> * We fix the validation directly on the existing template - VM is affected

Existing VMs will not be functionally affected unless you try editing them again, right? Change in the template validations does not propagate to the VM.spec automatically. The template validator might start raising warnings though.

> I think the latter is something we should never do, because the VM could have been edited by the user by the time the fix was shipped and we might retroactively break existing workloads

Lets discuss what process is the right one:

I would argue that a bug in validations like allowing a wrong disk type or not enough memory for a given OS is a rather bad bug. When such a VM is created with the wrong settings it might not even boot up despite passing the validation. Fixing this should not have any negative impact on the VM then. Quite the opposite, it will tell the user what he did wrong and how to fix the VM.

The case you are describing is basically a too strict validation as the VM can run even when the values are not in the allowed range. So technically we could argue that in this case the validation rule should not be made more strict (less strict rather).

But there is also the legal aspect of things. And legally this has impact on the supportability. Just because the VM can run with non-supported values does not mean we want users to run the VMs like that. Maybe it will generate more support requests because of corner case bugs?
 
> I am unsure about safely upgrading VMs from template version X to template version X+1 (in case of a bugfix or a feature update), but I do not think this should be an automatic action, I think the user should actively confirm the upgrade and understand the circumstances, instead of us potentially breaking their workloads.

I agree. Upgrade to a new template version should always be something the user approves explicitly as the spec might change.

omeryahud:
> I would argue that a bug in validations like allowing a wrong disk type or not enough memory for a given OS is a rather bad bug. When such a VM is created with the wrong settings it might not even boot up despite passing the validation. Fixing this should not have any negative impact on the VM then. Quite the opposite, it will tell the user what he did wrong and how to fix the VM.

The case of a freshly processed VM not booting is of course a bad bug, but it is also an easy one to solve, a new template is provided, and the user is requested to re-process the VM.

I am worried about a VM that passed (bad) validations, and is running. I wouldn't want to interrupt the lifecycle of that VM once it has been accepted and is operational (although not optimal), that is my reasoning to treat validations/spec the same way, meaning neither of them should automatically propagate to already created VMs.

> But there is also the legal aspect of things. And legally this has impact on the supportability. Just because the VM can run with non-supported values does not mean we want users to run the VMs like that. Maybe it will generate more support requests because of corner case bugs?

If the user has issues with that VM down the road, we can document how they can manually migrate to the newest template version (until we provide code that does that for them): updating the spec according to the newest template, and copying over validation rules. If they still have issues, then we should anyway support them, as they are using the latest template provided by us.

jean-edouard:
@MarSik are your concerns resolved?
/lgtm

omeryahud:
/hold
This PR is should be merged after this TV PR (https://github.com/kubevirt/kubevirt-template-validator/pull/53) is merged and a new TV release is created

PR #292, opened 2020-12-03
use rtl8139 interface model on rhel6 and centos6

dankenigsberg:
rhel6 does not support the virtio interface model, which is kubevirt's
default. This change sets the interface model to rtl8139 explicitly on
rhel6, which is known to work there. It drops networkInterfaceMultiqueue
as it has no virtio device to apply on.

Bug-URL: https://bugzilla.redhat.com/show_bug.cgi?id=1794243
Signed-off-by: Dan Kenigsberg <danken@redhat.com>

**Release note**:

```release-note
VMs created based on the rhel6 or centos6 templates can communicate over the pod network with no need to select non-virtio model.
```

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -102,7 +102,6 @@ objects:
               memory: {{ item.memsize }}
           devices:
             rng: {}
-            networkInterfaceMultiqueue: true
 {% if item.tablet %}
             inputs:
               - type: tablet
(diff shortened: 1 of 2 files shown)


dankenigsberg:
centos lane fails with
```
 The request is invalid: spec.template.spec.domain.devices.networkInterfaceMultiqueue: virtio-net multiqueue request, but there are no virtio interfaces defined 
```
I need to disable networkInterfaceMultiqueue on rhel6 or remove this validation from KubeVirt.

Frankly, I think that the latter is better. `networkInterfaceMultiqueue` already means "enable multiqueue when it makes sense".

dankenigsberg:
/retest

```
 Unable to connect to the server: dial tcp 104.196.38.75:6443: i/o timeout 
```

PR #295, opened 2020-12-20
Explain why Windows grace period is so long

dankenigsberg:
Explain why Windows grace period is so long

commit 7127e2fd0 has set terminationGracePeriodSeconds for Windows
template on 3600 seconds, or one whole hour. I find it way too long, so
document the reason behind it. 

Signed-off-by: Dan Kenigsberg <danken@redhat.com>

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -284,7 +284,9 @@ status:
 ```
 
 ## terminationGracePeriodSeconds
-All linux templates have terminationGracePeriodSeconds set to 180 seconds, in windows templates it is set to 3600 seconds.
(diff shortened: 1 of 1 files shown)


rnetser:
The value was the recommendation as Windows may run updates or any other processes prior to shutting down and you do not want to kill the VMI while doing so. It is up to an hour; if the OS is ready to shut down, it will (before reaching the gracePeriod) and the VM will be stopped.
I think that setting it to 5 minutes is risky.

dankenigsberg:
ok @rnetser, I'm changing this to a doc PR.

PR #298, opened 2021-01-06
validate pvc name and namespace stability across upgrades

kwiesmueller:
Adds a test to ensure `SRC_PVC_NAME` and `SRC_PVC_NAMESPACE` do not change across upgrades.

Right now the test relies on the template names not changing across versions. If a template gets renamed the test won't fail as it's technically not the same template anymore.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/unit-tests.sh
+++ b/automation/unit-tests.sh
@@ -43,6 +43,15 @@ LATEST_CT=$(curl -L https://api.github.com/repos/kubevirt/common-templates/relea
             jq '.[] | select(.prerelease==false) | .name' | sort -V | tail -n1 | tr -d '"')
 oc apply -f https://github.com/kubevirt/common-templates/releases/download/${LATEST_CT}/common-templates-${LATEST_CT}.yaml
 
+set +e
+python3 automation/validate-pvc-name-stability.py
+RC=${?}
+set -e
+if [ ${RC} -ne 0 ];then
+  echo "[Upgrade] Validation of PVC name stability failed"
+  exit ${RC}
+fi
+
 oc apply -f dist/templates
 
 ./automation/test_duplicate_templates.sh
--- a/automation/validate-pvc-name-stability.py
+++ b/automation/validate-pvc-name-stability.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+import logging
+import os
+import os.path
+import yaml
+import sys
+
+from kubernetes import client, config
+from openshift.dynamic import DynamicClient
+
+
+def validatePVCNames(path, liveTemplates):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    errors = []
+    for templateFilename in templates:
+        with open(os.path.join(path, templateFilename), 'r') as stream:
+            template = yaml.safe_load(stream)
+
+            if template is None:
+                errors.append("Empty template file: {}".format(templateFilename))
+                continue
+
+            templateName = template["metadata"]["name"]
+            logging.info("Checking PVC name stability for: {}".format(templateName))
+
+            pvcName = getPVCNameFrom(template)
+            pvcNamespace = getPVCNamespaceFrom(template)
+
+            matchingLiveTemplate = liveTemplates.get(templateName)
+            if matchingLiveTemplate:
+                liveTemplatePVCName = getPVCNameFrom(matchingLiveTemplate)
+                liveTemplatePVCNamespace = getPVCNamespaceFrom(matchingLiveTemplate)
+
+                if pvcName != liveTemplatePVCName:
+                    errors.append("PVC name: {} was modified in: {}".format(pvcName, templateName))
+                if pvcNamespace != liveTemplatePVCNamespace:
+                    errors.append("PVC namespace: {} was modified in: {}".format(pvcNamespace, templateName))
+            else:
+                logging.info("Missing liveTemplate for {}".format(templateName))
+
+    if errors:
+        error_message = "\n".join(errors)
+        logging.info("PVC stability vaidation failed")
+        raise Exception(error_message)
+
+
+def getParamFrom(template, paramName):
+    for param in template["parameters"]:
+        if param["name"] == paramName:
+            return param["value"]
+    return None
+
+
+def getPVCNameFrom(template):
+    return getParamFrom(template, "SRC_PVC_NAME")
+
+
+def getPVCNamespaceFrom(template):
+    return getParamFrom(template, "SRC_PVC_NAMESPACE")
+
+
+def fetchLiveTemplates():
+    k8s_client = config.new_client_from_config()
+    dyn_client = DynamicClient(k8s_client)
+
+    template_v1 = dyn_client.resources.get(api_version='template.openshift.io/v1', kind='Template')
+    templates = template_v1.get()
+
+    if not templates:
+        return {}
+
+    liveTemplates = {}
+    for template in templates.items:
+        liveTemplates[template["metadata"]["name"]] = template
+
+    return liveTemplates
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
+    logging.info("Running PVC name stability validation")
+
+    try:
+        liveTemplates = fetchLiveTemplates()
+        validatePVCNames("dist/templates", liveTemplates)
+    except Exception as e:
+        logging.error(e)
+        sys.exit(1)


automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,103 @@
...
+def getPVCNameFrom(template):
+    return getParamFrom(template, "SRC_PVC_NAME")
+
+
+def getPVCNamespaceFrom(template):
+    return getParamFrom(template, "SRC_PVC_NAMESPACE")
+
+
+def fetchLiveTemplates():
+    out = subprocess.run(["oc", "get", "template", "-o", "yaml"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
ksimon1, line 66: what about using python openshift package? https://github.com/openshift/openshift-restclient-python

kwiesmueller: Oh, cool. I didn't know that exists.
I thought about adding the kubernetes client and using that, but decided to avoid any additional dependencies for the test.
Didn't want to add any python requirements as I didn't see any managed right now. But I can use the client if you'd prefer that.

ksimon1: We already have python dependencies (see https://github.com/kubevirt/common-templates/blob/master/builder/Dockerfile#L4). Please if you have time try to use the python client.

kwiesmueller: Done, thanks for pointing me in the right direction.

ksimon1:
/lgtm cancell
the test is failing:
```
 + python3 automation/validate-pvc-name-stability.py
Traceback (most recent call last):
  File "/go/src/github.com/kubevirt/common-templates/automation/validate-pvc-name-stability.py", line 10, in <module>
    from kubernetes import client, config
ModuleNotFoundError: No module named 'kubernetes'
make: *** [makefile:33: unit-tests] Error 1 
```

ksimon1:
/lgtm cancel
the test is failing:
```
 + python3 automation/validate-pvc-name-stability.py
Traceback (most recent call last):
  File "/go/src/github.com/kubevirt/common-templates/automation/validate-pvc-name-stability.py", line 10, in <module>
    from kubernetes import client, config
ModuleNotFoundError: No module named 'kubernetes'
make: *** [makefile:33: unit-tests] Error 1 
```

kwiesmueller:
Hm, test fails with a missing dependency. Do I need to do something else than adding the package to the Dockerfile?
Locally everything was properly installed.

ksimon1:
I think there is a problem, that openshift ci can't load the newest dockerfile. I encountered the similar problems during migration from prow to openshift ci. Can you please separate this PR to one, which contains the test and the second which has the dockerfile change?

kwiesmueller:
Let's see if that helps: https://github.com/kubevirt/common-templates/pull/308

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+
+    template_v1 = dyn_client.resources.get(api_version='template.openshift.io/v1', kind='Template')
+    templates = template_v1.get()
+
+    if not templates:
+        return {}
+
+    liveTemplates = {}
+    for template in templates.items:
+        liveTemplates[template["metadata"]["name"]] = template
```
akrejcir, line 81: There may be a template with the same name in different namespaces, in that case only one of them will be in the dictionary. To fix it, you could use a tuple as the key:
`(namespace, name)`.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
+#! python
+
+import logging
+import os
+import os.path
+import yaml
+import subprocess
```
akrejcir, line 7: The `subprocess` module is not used anywhere.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+
+def validatePVCNames(path, liveTemplates):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    errors = []
+    for templateFilename in templates:
+        with open(os.path.join(path, templateFilename), 'r') as stream:
+            try:
+                template = yaml.safe_load(stream)
+
+                if template == None:
```
akrejcir, line 22: Comparison with `None` should use `is`: `if template is None:`

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+    if len(errors) > 0:
+        error_message = "\n".join(errors)
+        logging.info(FAILED_INFO_MESSAGE)
+        raise Exception(error_message)
+
+
+def getParamFrom(template, paramName):
+    for param in template["parameters"]:
+        if param["name"] == paramName:
+            return param["value"]
```
akrejcir, line 58: If this function does not find the parameter `paramName`, it implicitly returns `None`.
It it is expected, please add an explicit `return None` at the end, to make it obvious.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+from kubernetes import client, config
+from openshift.dynamic import DynamicClient
+
+
+def validatePVCNames(path, liveTemplates):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    errors = []
+    for templateFilename in templates:
+        with open(os.path.join(path, templateFilename), 'r') as stream:
+            try:
```
akrejcir, line 19: Can you make this `try` block smaller, so it is only around statements that can throw exceptions? I think, it is probably only `template = yaml.safe_load(stream)`

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+                    logging.info("Empty template file: %s", template)
+                    continue
+
+                templateName = template["metadata"]["name"]
+                logging.info("Checking PVC name stability for: {}".format(templateName))
+
+                pvcName = getPVCNameFrom(template)
+                pvcNamespace = getPVCNamespaceFrom(template)
+
+                matchingLiveTemplate = liveTemplates.get(templateName, None)
```
akrejcir, line 32: The second argument to `get()` is `None` by default, so you don't need to write it explicitly.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+        liveTemplates[template["metadata"]["name"]] = template
+
+    return liveTemplates
+
+
+def pvcNameErrorMessage(pvcName, templateName):
+    return "PVC name: {} was modified in: {}".format(pvcName, templateName)
+
+
+def pvcNamespaceErrorMessage(pvcNamespace, templateName):
```
akrejcir, line 90: These two functions can be inlined. They are only one line and used only once.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,105 @@
...
+    return "PVC namespace: {} was modified in: {}".format(pvcNamespace, templateName)
+
+
+FAILED_INFO_MESSAGE = "PVC stability vaidation failed"
```
akrejcir, line 94: None of the other log messages use a variable, so this one can be inlined too.

akrejcir (changes requested):
The logic is mostly good, I only have one comment about using namespaces in dictionary key.
Other comments are mostly about code style, you don't have to fix them if you think it is not necessary. 

Can you also squash the commits? They are related.

akrejcir:
Can you remove the `Signed-off-by:...` from the PR's description?

automation/validate-pvc-name-stability.py
kwiesmueller: Right now it only fetches templates from one namespace.
Should it take them from different namespaces? As I thought all our common-templates land in one.
If yes, I agree with you.

omeryahud: In the context of the tests, we only deploy the templates in a single namespace, so I don't mind if it stays this way, please decide what you intend to do so we can approve this

kwiesmueller: I would keep it like it is then.

automation/unit-tests.sh
```diff
@@ -43,6 +43,12 @@ LATEST_CT=$(curl -L https://api.github.com/repos/kubevirt/common-templates/relea
             jq '.[] | select(.prerelease==false) | .name' | sort -V | tail -n1 | tr -d '"')
 oc apply -f https://github.com/kubevirt/common-templates/releases/download/${LATEST_CT}/common-templates-${LATEST_CT}.yaml
 
+python3 automation/validate-pvc-name-stability.py
+if [ $? -eq 1 ];then
```
sarahbx, line 47: A normal exit should return `0`, so may want to use `-ne 0` instead.

automation/unit-tests.sh
```diff
@@ -43,6 +43,12 @@ LATEST_CT=$(curl -L https://api.github.com/repos/kubevirt/common-templates/relea
...
+  echo "[Upgrade] Validation of PVC name stability failed"
+  exit 1
```
sarahbx, line 49: Related to previous note: If we will exit on any non-zero, we should pass that value through here so it may be consumed by the caller.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def validatePVCNames(path, liveTemplates):
+    templates = [f for f in os.listdir(path) if os.path.isfile(os.path.join(path, f))]
+    errors = []
+    for templateFilename in templates:
+        with open(os.path.join(path, templateFilename), 'r') as stream:
+            try:
+                template = yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
```
sarahbx, line 20: Would it be possible to add a else case to the try/except? Even just to record that we hit a uncaught error.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,92 @@
...
+                liveTemplatePVCNamespace = getPVCNamespaceFrom(matchingLiveTemplate)
+
+                if pvcName != liveTemplatePVCName:
+                    errors.append("PVC name: {} was modified in: {}".format(pvcName, templateName))
+                if pvcNamespace != liveTemplatePVCNamespace:
+                    errors.append("PVC namespace: {} was modified in: {}".format(pvcNamespace, templateName))
+            else:
+                logging.info("Missing liveTemplate for {}".format(templateName))
+
+    if len(errors) > 0:
```
sarahbx, line 45: This could be reduced to `if errors:` unless we want to also check the instance type.

automation/validate-pvc-name-stability.py
sarahbx, line 20: We may want to add an `else` case here so causes of other errors are recorded also.

automation/unit-tests.sh
kwiesmueller: Do the two comments also apply to the other uses inside this file?
Should I change them as well?

automation/validate-pvc-name-stability.py
kwiesmueller: That's what you get from mostly only doing go for 4 years.

automation/validate-pvc-name-stability.py
```diff
@@ -0,0 +1,94 @@
+#! python
```
sarahbx, line 1: The shebang should specify an absolute path. It is recommended to use `env` to locate python for portability. eg: `#!/usr/bin/env python`

kwiesmueller:
Alright, builds are green again.
Could you have another look @sarahbx ?

PR #300, opened 2021-01-07
add support for scsi disk bus in windows validation 

ksimon1:
When VMIO migrates a VM from RHV, it needs to be able to set the disk bus to "scsi".

This commit adds support for scsi in windows validation.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1912908

**Release note**:
```release-note
add support for scsi disk bus in windows validation 
```

### DIFF
--- a/templates/windows2k16.tpl.yaml
+++ b/templates/windows2k16.tpl.yaml
@@ -67,8 +67,8 @@ objects:
             "path": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
             "valid": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
             "rule": "enum",
-            "message": "disk bus has to be either virtio or sata",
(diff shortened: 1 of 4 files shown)


templates/windows2k16.tpl.yaml
```diff
@@ -68,7 +68,7 @@ objects:
             "valid": "jsonpath::.spec.domain.devices.disks[*].disk.bus",
             "rule": "enum",
             "message": "disk bus has to be either virtio or sata",
```
omeryahud, line 70: The rule message should be updated as well

omeryahud (changes requested):

templates/windows2k16.tpl.yaml
ksimon1: fixed

PR #302, opened 2021-01-10
Adopt RHEL6 and Centos 6 templates to new libvirt/QEMU changes

enp0s3:
Old guests like RHEL6 or CentOS6 require virtio-transitional
in newer libvirt/qemu versions, therefore a flag was added in the VM
template in order to signal the need for the virtio-transitional
configuration in the devices section.

Signed-off-by: Igor Bezukh <ibezukh@redhat.com>

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=1783192

**Release note**:
```release-note
NONE
```

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -101,6 +101,7 @@ objects:
             requests:
               memory: {{ item.memsize }}
           devices:
+            useVirtioTransitional: true
             rng: {}
 {% if item.tablet %}
             inputs:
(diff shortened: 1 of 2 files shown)


omeryahud:
Shouldn't we now encourage the use of virtio now that we can support it with RHEL/CentOS 6?
If so, the disk bus and NIC model should be updated as well.
@rmohr WDYT?

enp0s3:
> Shouldn't we now encourage the use of virtio now that we can support it with RHEL/CentOS 6?
> If so, the disk bus and NIC model should be updated as well.
> @rmohr WDYT?

@omeryahud I would suggest to create a separate PR for that because it has another context. The aim of this PR is just to provide a fix for the current situation.

PR #305, opened 2021-01-11
Centos 6 and RHEL 6 to use virtio bus for network and storage by default

enp0s3:
Old guests can now use virtio bus for disks and network interfaces
therefore configure the template to use that bus type as a default bus for them.

Signed-off-by: Igor Bezukh <ibezukh@redhat.com>

**Release note**:
```release-note
Centos 6 and RHEL 6 based images to consume virtio bus for disks and network devices
```

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -1,4 +1,4 @@
-{% set diskbus = diskbus |default("sata") %}
+{% set diskbus = diskbus |default("virtio") %}
 apiVersion: template.openshift.io/v1
 kind: Template
 metadata:
@@ -119,7 +119,7 @@ objects:
             interfaces:
             - masquerade: {}
               name: default
-              model: e1000e
+              model: virtio
         terminationGracePeriodSeconds: 180
         networks:
         - name: default
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -101,6 +101,7 @@ objects:
             requests:
               memory: {{ item.memsize }}
           devices:
+            networkInterfaceMultiqueue: true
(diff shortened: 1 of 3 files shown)


templates/centos6.tpl.yaml
```diff
@@ -119,7 +119,7 @@ objects:
             interfaces:
             - masquerade: {}
               name: default
-              model: e1000e
+              model: virtio
```
ksimon1, line 122: Shouldn't be this change in disk section? Because now you are updating network interface

enp0s3: @ksimon1 I've updated the PR subject and description. The aim of this PR is to swtich both disk and network devices to work with virtio.

enp0s3:
/hold
Need to validate with latest centos 6 image on plain kubevirt

dankenigsberg:
As this changes user-facing behavior, a release note is needed.

Given virtio is back, I suggest to reintroduce  `networkInterfaceMultiqueue: true`

enp0s3:
> As this changes user-facing behavior, a release note is needed.
> 
> Given virtio is back, I suggest to reintroduce `networkInterfaceMultiqueue: true`

@dankenigsberg Done

enp0s3:
/retest
the fix was merged to KV release 0.37

enp0s3:
/retest
```
export KUBEVIRT_VERSION=$(_curl -L https://api.github.com/repos/kubevirt/kubevirt/releases | \
            jq '.[] | select(.prerelease==false) | .name' | sort -V | tail -n1 | tr -d '"')
```
Release tag ready and I've verified that now ```KUBEVIRT_VERSION``` is  ```v0.37.2```

PR #319, opened 2021-02-02
osinfo-db: Update to tag v20210202

fidencio:
For the distros KubeVirt has a template, this update introduces the
following changes:
- Adds RHEL-8.4 as a pre-release
- Adjusts Ubuntu 19.10 tree URLs
- Updates Debian 9 & 10 URLs
- Adds data about CentOS 8.3.2011
- Adjusts CentOS 6.10 URLs after its EOL


fidencio:
@omeryahud, please, take a look.

ksimon1:
@fidencio can you please update your PR and keep the `osinfo-db-override` folder and just remove `os/redhat.com/rhel-8.1.xml` subfolders?

fidencio:
@ksimon1, a folder in git is automatically removed if empty, which is the case of `osinfo-db-override`.  May I ask what broke? If you can point me to the part that broke, I may be able to provide a patch to check whether the directory exists before accessing it, and that would be the correct way to go.

Does this sound reasonable?

ksimon1:
The tests are failing on https://github.com/kubevirt/common-templates/blob/master/lookup_plugins/osinfo.py#L31
So if you can please update it to not fail, when the folder does not exist.

fidencio:
@ksimon1, please, take a look at the up-to-date series.

fidencio:
@ksimon1, Let me close this PR as I'll do a new release Today, including RHEL-9 pre-release, and update it altogether here.

PR #297, opened 2021-01-05
[WIP] Debug PR for #296

nelsonspbr:
**What this PR does / why we need it**: This PR adds debug messages to `automation/test-linux.sh` on top of the modifications proposed in #296. This should help debugging E2E tests that are failing at the moment. For now this is a WIP / not-to-be-merged PR.

**Release note**:

```release-note
NONE
```

PR #323, opened 2021-02-10
Add support for multiqueue networking for high performance Windows templates

nelsonspbr:

This PR add support for multiqueue networking for Windows VMs. This requires two modifications:
- set `vmi.spec.domain.devices.networkInterfaceMultiqueue` appropriately; and
- set `vmi.spec.domain.devices.interfaces.model` to `virtio`.

The current proposal adds a template flag, `multiqueue` which is set to true for high performance Windows templates.
This flag is then processed in each template file to set the correct value for `devices.networkInterfaceMultiqueue` and `devices.interfaces.model`.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/windows10.tpl.yaml
+++ b/templates/windows10.tpl.yaml
@@ -144,13 +144,20 @@ objects:
               runtime: {}
               reset: {}
           devices:
+{% if item.multiqueue %}
+            networkInterfaceMultiqueue: True
+{% endif %}
             disks:
             - disk:
                 bus: sata
               name: ${NAME}
             interfaces:
             - masquerade: {}
+{% if item.multiqueue %}
+              model: virtio
+{% else %}
               model: e1000e
+{% endif %}
               name: default
 {% if item.tablet %}
             inputs:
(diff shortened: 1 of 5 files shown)


templates/windows10.tpl.yaml
```diff
@@ -144,13 +144,20 @@ objects:
...
+            networkInterfaceMultiqueue: True
+{% endif %}
             disks:
             - disk:
                 bus: sata
               name: ${NAME}
             interfaces:
             - masquerade: {}
+{% if item.multiqueue %}
+              model: virtio
```
ksimon1, line 157: Btw. don't need virtio drivers to be installed in windows image?

nelsonspbr:
Added debug information similar to what was done [here](https://github.com/kubevirt/common-templates/pull/297/commits/4ff2f93b74dc946cded595305efe1717aa3f3aa6). Debug code will be removed later.

templates/windows10.tpl.yaml
nelsonspbr: Ah, I thought they were already in these images we use. I will keep an eye on it, especially for the Windows 10 test.

Regarding users, the idea is to require them to have it set up properly before using the high performance templates.

nelsonspbr:
All tests passed. Removed debug code and running a final set of tests.

nelsonspbr:
/retitle Add support for multiqueue networking for high performance Windows templates

PR #325, opened 2021-02-15
roll back to previous version version of CDI

ksimon1:
roll back to previous version version of CDI

newest version contain bug, which prevent VM to run

**Release note**:

```
NONE
```


kwiesmueller:
Linking https://github.com/kubevirt/containerized-data-importer/pull/1659 here as we should revert this once the fix is available.

PR #330, opened 2021-02-18
Improve tests stability

ksimon1:

tests don't have to fail during VM creation
If vm is not created due to some cluster issue, test doesn't have to fail,
beacuse the retry feature can create vm again.

Command which allows cpumanager on the nodes moved to test.sh script,
because it doesn't have to run everytime vm is created.

//cc @omeryahud 

**Release note**:
```
NONE
```


ksimon1:
@omeryahud, @nelsonspbr, @akrejcir can you please review? This should fix our problems with tests.

PR #351, opened 2021-04-19
Periodic job for creating a container disk image for the latest CentOS Version

shwetaap:
Periodic job for creating a container disk image for the latest CentOS Version

**Release note**:
```release-note
None
```

Signed-off-by: Shweta Padubidri <spadubid@redhat.com>


dvtemplates/new_cdi_image.sh
```diff
@@ -0,0 +1,102 @@
...
+    error_message 8
+fi
+
+FEDORA_OS="fedora"
+CENTOS="centos"
+
+# Set the Image registry and identify the latest version of the OS
+if [[ "${TARGET_OS}" == "${CENTOS}" ]]; then
+    cd ${PWD}/centos
+    OS_REPO="quay.io/shwetaap/centos-images"
```
ksimon1, line 42: can you please change it to kubevirt repo?

dvtemplates/new_cdi_image.sh
```diff
@@ -0,0 +1,102 @@
...
+re='^[0-9]+$'
+if ! [[ $OS_VERSION =~ $re ]] ; then
+    error_message 2
+fi
+echo "Latest ${TARGET_OS} version is : ${OS_VERSION}"
+
+# Fetch the old image
+docker pull -a $OS_REPO
+OS_OLD_VERSION=$(docker images $OS_REPO --format "{{json .}}" | jq 'select(.Tag|test("^[0-9]+$")) | .Tag' | sort -rn | head -n 1 | tr -d '"')
+echo "${TARGET_OS} version in the Image Registry is : ${OS_OLD_VERSION}"
```
ksimon1, line 62: Wouldn't be better to check hash of the image and then compare hashes?

shwetaap: At present,
The new version is got from the qcow2 image of the OS. 
The old version is got from using the container disk image(which has the qcow2 image within it). 
So one of the images is qcow2 image, and the other is a container disk image. So we cannot do a direct image hash comparison.

dvtemplates/new_cdi_image.sh
shwetaap: Done

dvtemplates/build_k8s_cluster.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Build a K8s cluster from kubevirtci
```
sarahbx, line 3: Please `set -e` in all shell scrips so default is to exit on error. If you need to allow a command to fail, we should address those individually.

dvtemplates/build_k8s_cluster.sh
```diff
@@ -0,0 +1,20 @@
...
+cd ${PWD}/../kubevirtci
```
sarahbx, line 4: Please put any strings that have variables in them within double quotes to prevent possible word splitting.

dvtemplates/new_cdi_image.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+error_message() {
+    OS=$2
+    if [ $1 -eq 1 ]; then
+        echo -e "Unable to access data from the OS release URL"
```
sarahbx, line 6: Can we assign these integers to named constants? Would help readability when used.

dvtemplates/new_cdi_image.sh
```diff
@@ -0,0 +1,102 @@
...
+
+if [[ "$OS_VERSION" == "$OS_OLD_VERSION" ]]; then
+    error_message 4 "$TARGET_OS"
+fi
+
+#Download the latest OS Image
+if [[ "${TARGET_OS}" == "${CENTOS}" ]]; then
+    IMAGE_URL=${BASE_URL}${OS_VERSION}/x86_64/images/
+    wget -qO image.html $IMAGE_URL/CHECKSUM || exit 3
+    CentOS_IMAGE=$(grep ^#.*GenericCloud.*qcow2 image.html | sed -e 's/.*\(CentOS.*qcow2\).*/\1/')
```
sarahbx, line 76: Please enclose the regex input to grep within quotes so it will not be interpreted by bash.

sarahbx (changes requested):

dvtemplates/build_k8s_cluster.sh
shwetaap: Done

dvtemplates/build_k8s_cluster.sh
shwetaap: Done

dvtemplates/new_cdi_image.sh
shwetaap: Done

dvtemplates/new_cdi_image.sh
shwetaap: Done

dvtemplates/build_k8s_cluster.sh
sarahbx: 👍 We're going to want to remove the `x` on these before merge, to prevent exposing any sensitive data.

dvtemplates/build_k8s_cluster.sh
sarahbx: May just want to put the entire string within the quotes so it's clear it's a single value.
eg: `cd "${PWD}/../kubevirtci"`

dvtemplates/build_k8s_cluster.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -ex
```
sarahbx, line 3: We're going to want to remove the x on these, to prevent exposing any sensitive data.

dvtemplates/build_k8s_cluster.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Build a K8s cluster from kubevirtci
+cd "${PWD}"/../kubevirtci
```
sarahbx, line 6: May just want to put the entire string within the quotes so it's clear it's a single value.
eg: `cd "${PWD}/../kubevirtci"`

dvtemplates/build_k8s_cluster.sh
shwetaap: Done

dvtemplates/build_k8s_cluster.sh
shwetaap: Done

dvtemplates/new_cdi_image.sh
ksimon1: But you can do it like: download new qcow2 image and calculate hash, put this hash as a label during docker build. Then download old image from quay and check its hash label. And then you can easily compare if the hashes are identical.

shwetaap: Yes that can be done. Thanks! Will add it as a different patch to improve the existing working functionalitly.

dvtemplates/new_cdi_image.sh
```diff
@@ -0,0 +1,115 @@
...
+    error_message "$ERROR_MISSING_OS_IMAGE"
+fi
+
+FEDORA_OS="fedora"
+CENTOS="centos8"
+
+# Set the Image registry and identify the latest version of the OS
+if [[ "${TARGET_OS}" == "${CENTOS}" ]]; then
+    cd "${PWD}/centos8"
+    OS_REPO="quay.io/kubevirt/centos8-container-disk-images"
```
kwiesmueller, line 54: Are the two repos of a different naming schema on purpose?

dvtemplates/new_cdi_image.sh
kwiesmueller: Please do this as a follow-up PR. Otherwise
/lgtm
/approve

dvtemplates/new_cdi_image.sh
shwetaap: Yea, I thought this name made more sense. I can check with Federico and update the Fedora repo in a follow-up PR as well.

PR #357, opened 2021-04-27
Disable testing of high performance templates.

ksimon1:
Disable testing of high performance templates.
Due to missing cpu manager on testing nodes, we have to disable testing
of HP templates

**Release note**:
```
NONE
```


kwiesmueller:
/lgtm
Let's clarify that this is a workaround to unblock CI until the underlying issue is fixed.

PR #359, opened 2021-04-28
allow testing of HP templates

ksimon1:
allow testing of HP templates
Enable CPU manager on ci nodes

The cpu manager is allowed only for templates, which have high performance variant. Other templates don't need it, because they don't have HP variant. I did it this way, because it takes about 10 minutes to enable cpu manager on nodes.

**Release note**:
```
NONE
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -17,6 +17,29 @@
 # Copyright 2018 Red Hat, Inc.
 #
 
+# Start CPU manager only for templates which require it.
+if [[ $TARGET =~ rhel7.* ]] || [[ $TARGET =~ rhel8.* ]] || [[ $TARGET =~ fedora.* ]] || [[ $TARGET =~ windows2.* ]]; then
+  oc label machineconfigpool worker custom-kubelet=enabled
+  oc create -f - <<EOF
+---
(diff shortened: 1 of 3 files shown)


automation/test.sh
```diff
@@ -17,6 +17,29 @@
 # Copyright 2018 Red Hat, Inc.
 #
 
+# Start CPU manager only for templates which require it.
```
akrejcir, line 20: Why is the CPU manager started only for certain templates? Is it only a performance optimization, so the tests take less time? If yes, how much slower are tests with CPU manager?

I don't like that there is a list of template names in this test script. If the performance hit is not too big, the CPU manager could be enabled for all templates.

ksimon1: The cpu manager is allowed only for templates, which have high performance variant. Other templates don't need it, because they don't have HP variant. I did it this way, because it takes about 10 minutes to enable cpu manager on nodes.

akrejcir: Ok, that makes sense.
Can you write this information to the comment? So it is clear why is it needed only for these templates. And maybe can you also mention the specific feature of HP VMs that requires the CPU manager?

ksimon1: done

PR #363, opened 2021-06-10
Support VMs with legacy i440fx machine type

vladikr:
This PR provides an RHEL6 based meta-template, to support legacy guest OSes 
which require an older pc-i440fx machine type to boot

**Release note**:
```release-note
NONE
```

### DIFF
--- a/templates/contrib/pc/README.md
+++ b/templates/contrib/pc/README.md
@@ -1,0 +1,23 @@
+# Overview
+
+This directory provides a RHEL6 based template, to support legacy guest OS
+which require an older pc-i440fx machine type to boot.
+Clusters that were deployed using the HyperConverged Operator do not include
+support for this machine type. To allow Virtual Machines to request this
+machine type, this option should be manually added.
+
+# Update HCO
+
+It is necessary to add the relevant machine type to a list of permitted types.
+This will allow Virtual Machines to request the pc-i440fx machine type.
+
+$ oc annotate --overwrite -n openshift-cnv hco kubevirt-hyperconverged \
+  kubevirt.kubevirt.io/jsonpatch='[{"op": "add", \
+    "path": "/spec/configuration/emulatedMachines", \
+    "value": ["q35*", "pc-q35*", "pc-i440fx-rhel7.6.0"] }]'
+
+# Usage
+
+The provided template can be used to create a Virtual Machine
+
+$ oc process --local -f templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
--- a/templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
+++ b/templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
@@ -1,0 +1,155 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: rhel6-server-large-pc-i440fx
+  annotations:
+    openshift.io/display-name: "Red Hat Enterprise Linux 6.0+ VM with i440fx machine type"
+    description: >-
+      Template for Red Hat Enterprise Linux 6 VM or newer.
+      A PVC with the RHEL disk image must be available.
+    tags: "hidden,kubevirt,virtualmachine,linux,rhel"
+    iconClass: "icon-rhel"
+    openshift.io/provider-display-name: "KubeVirt"
+    openshift.io/documentation-url: "https://github.com/kubevirt/common-templates"
+    openshift.io/support-url: "https://github.com/kubevirt/common-templates/issues"
+    template.openshift.io/bindable: "false"
+    template.kubevirt.io/version: v1alpha1
+    defaults.template.kubevirt.io/disk: rootdisk
+    template.kubevirt.io/editable: |
+      /objects[0].spec.template.spec.domain.cpu.sockets
+      /objects[0].spec.template.spec.domain.cpu.cores
+      /objects[0].spec.template.spec.domain.cpu.threads
+      /objects[0].spec.template.spec.domain.resources.requests.memory
+      /objects[0].spec.template.spec.domain.devices.disks
+      /objects[0].spec.template.spec.volumes
+      /objects[0].spec.template.spec.networks
+    name.os.template.kubevirt.io/rhel6.0: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.1: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.10: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.2: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.3: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.4: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.5: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.6: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.7: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.8: Red Hat Enterprise Linux 6.0 or higher
+    name.os.template.kubevirt.io/rhel6.9: Red Hat Enterprise Linux 6.0 or higher
+  labels:
+    os.template.kubevirt.io/rhel6.0: "true"
+    os.template.kubevirt.io/rhel6.1: "true"
+    os.template.kubevirt.io/rhel6.10: "true"
+    os.template.kubevirt.io/rhel6.2: "true"
+    os.template.kubevirt.io/rhel6.3: "true"
+    os.template.kubevirt.io/rhel6.4: "true"
+    os.template.kubevirt.io/rhel6.5: "true"
+    os.template.kubevirt.io/rhel6.6: "true"
+    os.template.kubevirt.io/rhel6.7: "true"
+    os.template.kubevirt.io/rhel6.8: "true"
+    os.template.kubevirt.io/rhel6.9: "true"
+    workload.template.kubevirt.io/server-pc-i440fx: "true"
+    flavor.template.kubevirt.io/large: "true"
+objects:
+- apiVersion: kubevirt.io/v1alpha3
+  kind: VirtualMachine
+  metadata:
+    name: ${NAME}
+    labels:
+      vm.kubevirt.io/template: rhel6-server-large-pc-i440fx
+      vm.kubevirt.io/template.version: "devel"
+      vm.kubevirt.io/template.revision: "1"
+      app: ${NAME}
+    annotations:
+      vm.kubevirt.io/validations: |
+        [
+          {
+            "name": "minimal-required-memory",
+            "path": "jsonpath::.spec.domain.resources.requests.memory",
+            "rule": "integer",
+            "message": "This VM requires more memory.",
+            "min": 536870912
+          }
+        ]
+  spec:
+    dataVolumeTemplates:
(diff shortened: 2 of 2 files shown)


generate-templates.yaml
```diff
@@ -121,6 +121,23 @@
...
       cloudusername: cloud-user
 
+  - name: Generate RHEL 6 templates with a legacy machine type pc-i440fx
+    template:
+      src: templates/contrib/pc/rhel6.tpl.yaml
+      dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}-pc.yaml"
+    with_items:
+    - {flavor: tiny,   workload: server,  memsize: "1Gi", cpus: 1, iothreads: False, emulatorthread: False, tablet: False, default: False}
+    - {flavor: small,  workload: server,  memsize: "2Gi", cpus: 1, iothreads: False, emulatorthread: False, tablet: False, default: True}
+    - {flavor: medium, workload: server,  memsize: "4Gi", cpus: 1, iothreads: False, emulatorthread: False, tablet: False, default: False}
```
fabiand, line 131: Please drop these 3 lines, we want to keep it simple and just have one template

generate-templates.yaml
```diff
@@ -121,6 +121,23 @@
       osinfoname: "{{ oslabels[0] }}"
       cloudusername: cloud-user
 
+  - name: Generate RHEL 6 templates with a legacy machine type pc-i440fx
+    template:
+      src: templates/contrib/pc/rhel6.tpl.yaml
```
fabiand, line 126: @ksimon1 we want this template, but we do not want to publish it as part of the releases, how can we do this?

ksimon1: Add command
`rm dist/templates/<Template name>`
here https://github.com/kubevirt/common-templates/blob/master/makefile#L26

templates/contrib/pc/README.md
```diff
@@ -0,0 +1,14 @@
+# Overview 
+
+This directory provides a RHEL6 based meta-template, to support legacy guest OS
```
ksimon1, line 3: I think you will have to change this description to something like:
1) create copy of rhel 6 template
2) update machine type on the new template
because if you change the something on original template, ssp operator will replace it back to its original state.

templates/contrib/pc/rhel6.tpl.yaml
```diff
@@ -0,0 +1,157 @@
+{% set diskbus = diskbus |default("virtio") %}
```
ksimon1, line 1: Do you plan to update this template with more changes? I would vote for reusing original rhel 6 jinja template and add `if` statement.

vladikr: Normally I would definitely follow your advice. 
In this case, I think, it would be better to explicitly make a copy and avoid introducing any changes in the original template to not hint that we support different machine types.
@fabiand wdyt?

templates/contrib/pc/README.md
vladikr: hmm, I think I'm not following your thought...
Isn't creating a separate meta-template address this issue?

templates/contrib/pc/rhel6.tpl.yaml
fabiand: I would also generally follow Karel's advice.
In thit special case I'd also however go with a copy, because we really want to avoid that somebody starts to have the idea that the machine type can be anything else than q35. Therefore a copy would be appreciated.
We could actually consider to make it a real modified copy - no ansible templating, copy of the final template with a modified machine type.

templates/contrib/pc/README.md
ksimon1: Sorry my fault, I wrongly understand the command bellow

templates/contrib/pc/rhel6.tpl.yaml
```diff
@@ -0,0 +1,157 @@
...
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
```
ksimon1, line 5: I think you should change name of this template, because if applied, it will replace regular rhel 6 template in cluster.
and currently this template is not tested. Can you please update tests here: https://github.com/kubevirt/common-templates/blob/master/automation/test-linux.sh

fabiand: Yes, let's just append `-pc` to the name

vladikr: That's done ;)

templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
```diff
@@ -0,0 +1,157 @@
...
+    os.template.kubevirt.io/rhel6.4: "true"
+    os.template.kubevirt.io/rhel6.5: "true"
+    os.template.kubevirt.io/rhel6.6: "true"
+    os.template.kubevirt.io/rhel6.7: "true"
+    os.template.kubevirt.io/rhel6.8: "true"
+    os.template.kubevirt.io/rhel6.9: "true"
+    workload.template.kubevirt.io/server-pc-i440fx: "true"
+    flavor.template.kubevirt.io/large: "true"
+    template.kubevirt.io/type: "base"
+    template.kubevirt.io/version: "devel"
```
akrejcir, line 52: These 2 labels need to be removed:
```
template.kubevirt.io/type: "base"
template.kubevirt.io/version: "devel"
```
Otherwise, the SSP operator will take ownership of the template, are remove it, because the `devel` version is different from the latest version deployed by the operator.

templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
```diff
@@ -0,0 +1,157 @@
...
+objects:
+- apiVersion: kubevirt.io/v1
```
dominikholler, line 54: Does this match the expected API version?

templates/contrib/pc/README.md
```diff
@@ -0,0 +1,23 @@
...
+# Update HCO
+
+It is necessary to add the relevant machine type to a list of permitted types.
+This will allow Virtual Machines to request the pc-i440fx machine type.
+
+$ oc annotate --overwrite -n openshift-cnv hco kubevirt-hyperconverged \
+  kubevirt.kubevirt.io/jsonpatch='[{"op": "add", \
+    "path": "/spec/configuration/emulatedMachines", \
+    "value": ["q35*", "pc-q35*", "pc*"] }]'
+
```
dominikholler, line 18: Is the updated annotation applied instantly, or is an additional step required?

fabiand: Does this expression also need ot be extended to permit i.e. a RHEL machine type?

vladikr: > Does this expression also need ot be extended to permit i.e. a RHEL machine type?

No, this doesn't need to be as rhel is a suffix. Unless we want to fixate on a specific one (pc-i440fx-rhel7.6.0), however, in that case, a slight change will lead to changes in the template.

templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
vladikr: Thanks. Will do.

templates/contrib/pc/README.md
vladikr: > Is the updated annotation applied instantly, or is an additional step required?

As far as I know, no additional steps are required.

vladikr: On second thought, I'll change this to match the template.

templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
dominikholler: On current downstream kubevirt-hyperconverged-operator.v2.6.5 I got
```
[dominik@t460p pc]$ oc process --local -f rhel6-server-large-pc-i440fx.yaml | oc apply -f -
unable to recognize no matches for kind "VirtualMachine" in version "kubevirt.io/v1"
```

vladikr: IIUC, this version should be fine.
It's also the version that appears in all templates in this repo.
For example: https://github.com/kubevirt/common-templates/blob/master/templates/rhel6.tpl.yaml#L42

omeryahud: I think `v1` virtual machines were introduced in CNV 4.8

templates/contrib/pc/README.md
vladikr: Just to double check... @nunnatsa would any additional steps be required here to change the `emulatedMachines` value in KubeVirt CR via the HCO CR?

templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
akrejcir: The `v1` version is available since kubevirt `v0.37.0`. 
If we want the template to be usable on older kubevirt, you can simply change the apiVersion:
```
apiVersion: kubevirt.io/v1alpha3
```
The template should be compatible with the older version without any other changes.

vladikr: somehow I was sure that we've introduced it in 0.36 and it's part of the 2.6 release.
But you are right, thank you for point this out.
I'll change it to `kubevirt.io/v1alpha3`

templates/contrib/pc/README.md
fabiand: @nunnatsa ?

dominikholler: I modified the machine-type of kubevirt-config ConfigMap to let kubevirt accept the new machine type.

templates/contrib/pc/rhel6-server-large-pc-i440fx.yaml
dominikholler: Thanks, works now on kubevirt-hyperconverged-operator.v2.6.5 . I don't have enough knowledge to understand the permission model of the dataVolumeTemplates source, but it might be fine.

akrejcir:
/lgtm

The changes in this PR are not tested by CI, so I'm overriding it.
/override ci/prow/e2e-azure-common-templates-windows2019

templates/contrib/pc/README.md
nunnatsa: I this is a formally required solution, I think it will be better to add it to HCO API. The annotation is for workarounds, solving urgent issues, and so on.

Maybe it will be better to apply a PR in HCO with a suggestion for a new API?

dominikholler: > Maybe it will be better to apply a PR in HCO with a suggestion for a new API?

Adding non-default machine types will stay an exception. For this reason, it might not worth an API change.

It might be helpful for the users, if we would explain in the documentation, how allow the new machine type not only with the upcoming release by the annotation, but also for the latest release 2.6 using the config map.

vladikr: > It might be helpful for the users, if we would explain in the documentation, how allow the new machine type not only with the upcoming release by the annotation, but also for the latest release 2.6 using the config map.

This is a little bit tricky, as the config will override any KubeVirt CR configuration. Therefore, all the configuration should be done either on the CR or on the config map.

@nunnatsa does the annotation work in the 2.6 release as well?

nunnatsa: > @nunnatsa does the annotation work in the 2.6 release as well?

No, it's not.

ksimon1:
/approve cancel
Cancelling approve, as this PR will be closed and created a new one

templates/contrib/pc/README.md
vladikr: OK, thanks. So what is the best approach to update this in 2.6? Should we suggest a config map or just update KubeVirt CR directly?

fabiand:
With the supplemental templates, can we close this?

vladikr:
Thanks, everyone. It was a good exercise ;)

fabiand:
The template landed at https://github.com/RHsyseng/cnv-supplemental-templates

PR #365, opened 2021-06-25
Validate that this repository does not contain offensive language

dominikholler:
Add a check to the sannity test that fails if there is a usage
of offensive language in one of the files.

The script forbid the use of the words `master`, `slave`,
`blacklist` and `whitelist`

Sinse `master` is very common word as a name of default
branches or as a node name in clusters, the script allows
the usage of it if this word is part of a URL or a path.
The script does not check the file
.github/workflows/release.yaml
that may contains a lot if usage of the word `master`, which
referes to the master branch.

@ksimon1 Can you help me to trigger this check on every PR?

**Release note**:
```
NONE
```

### DIFF
--- a/.github/workflows/validate-no-offensive-lang.yml
+++ b/.github/workflows/validate-no-offensive-lang.yml
@@ -1,0 +1,15 @@
+name: validate-no-offensive-lang
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: make validate-no-offensive-lang
+      run: make validate-no-offensive-lang
--- a/automation/validate-no-offensive-lang.sh
+++ b/automation/validate-no-offensive-lang.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+OFFENSIVE_WORDS="black[ -]?list|white[ -]?list|master|slave"
+ALLOW_LIST=".+[:/=]master[a-zA-Z]*/?"
+
+if git grep -inE "${OFFENSIVE_WORDS}" -- ":!${BASH_SOURCE[0]}" ':!.github/workflows/' | grep -viE "${ALLOW_LIST}"; then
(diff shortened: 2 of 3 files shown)


ksimon1:
You can trigger this script from two places. First is from makefile https://github.com/kubevirt/common-templates/blob/master/makefile#L32 and make it run with e.g. unit tests, or you can can create new test lane and run it completely separate (https://github.com/openshift/release/blob/master/ci-operator/config/kubevirt/common-templates/kubevirt-common-templates-master.yaml#L20) - I think this would be overkill and it can safely run with unit tests

dominikholler:
> You can trigger this script from two places. First is from makefile https://github.com/kubevirt/common-templates/blob/master/makefile#L32 and make it run with e.g. unit tests, or you can can create new test lane and run it completely separate (https://github.com/openshift/release/blob/master/ci-operator/config/kubevirt/common-templates/kubevirt-common-templates-master.yaml#L20) - I think this would be overkill and it can safely run with unit tests
>

Thanks, I added the new check as a dependency for the unit-tests.

dominikholler:
I added the action 'validate-no-offensive-lang / build (pull_request) '.

.github/workflows/validate-no-offensive-lang.yml
```diff
@@ -0,0 +1,15 @@
+name: validate-no-offensive-lang
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
kwiesmueller, line 10: Just checking if there is a fedora/centos environment available as that's what our own stuff runs in I believe.

kwiesmueller (commented):
One question, but
/lgtm
otherwise.

.github/workflows/validate-no-offensive-lang.yml
dominikholler: Good question, but unfortunately ubuntu is the only linux which is available on github hosted runners:  https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources .

PR #366, opened 2021-07-14
Add guest-agent ping based readiness probe

vatsalparekh:
**What this PR does / why we need it**:
Add a readiness probe to templates with `qemu-guest-agent` based `guest-ping`

**Special notes for your reviewer**:
Based on https://github.com/kubevirt/kubevirt/pull/5946 
**Release note**:

```release-note
NONE
```

PR #367, opened 2021-08-10
Add initial go test suite

vatsalparekh:
Add initial go test suite to run with https://github.com/openshift/release/pull/21019 

This is just the empty test suite right now so that #366 can directly go with adding its test on top of this and check it against a functest lane

**Release note**:

```release-note
NONE
```


akrejcir (changes requested):
Can you please add more information to the commit message? It should include the reason WHY this change is done.

akrejcir:
There is a typo in the commit message title.
Otherwise looks good.

akrejcir:
Looks good. Let's see if the CI passes.
We will need to change the offensive lang script to ignore vendor folder.

ksimon1:
can you please exclude the vendor folder from validate-no-offensive-lang script?

PR #374, opened 2021-08-31
switch pvc to storage in datavolume spec

ksimon1:
 Switch pvc to storage in datavolume spec
 
The storage type is similar to pvc but allows for some additional logic to be applied. It was introduced in order to implement detection and automation of storage parameters.

**Release note**:
```
 Switch pvc to storage in datavolume spec
```

### DIFF
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -66,9 +66,7 @@ objects:
       metadata:
         name: ${NAME}
       spec:
-        pvc:
-          accessModes:
-            - ReadWriteMany
+        storage:
           resources:
             requests:
               storage: 30Gi
--- a/automation/test-linux.sh
+++ b/automation/test-linux.sh
@@ -107,9 +107,7 @@ run_vm(){
   for i in `seq 1 3`; do
     error=false
     oc process ${template_option} -n $namespace -o json NAME=$vm_name SRC_PVC_NAME=$TARGET-datavolume-original SRC_PVC_NAMESPACE=kubevirt | \
-    jq 'del(.items[0].spec.dataVolumeTemplates[0].spec.pvc.accessModes) |
-    .items[0].spec.dataVolumeTemplates[0].spec.pvc+= {"accessModes": ["ReadWriteOnce"]} | 
-    .items[0].metadata.labels["vm.kubevirt.io/template.namespace"]="kubevirt"' | \
+    jq '.items[0].metadata.labels["vm.kubevirt.io/template.namespace"]="kubevirt"' | \
     oc apply -n $namespace -f -
 
     ./virtctl version
--- a/automation/test-windows.sh
+++ b/automation/test-windows.sh
@@ -86,9 +86,7 @@ run_vm(){
     error=false
 
     oc process -n $namespace -o json $template_name NAME=$vm_name SRC_PVC_NAME=$TARGET-datavolume-original SRC_PVC_NAMESPACE=kubevirt | \
-    jq 'del(.items[0].spec.dataVolumeTemplates[0].spec.pvc.accessModes) |
-    .items[0].spec.dataVolumeTemplates[0].spec.pvc+= {"accessModes": ["ReadWriteOnce"]} | 
-    .items[0].metadata.labels["vm.kubevirt.io/template.namespace"]="kubevirt"' | \
+    jq '.items[0].metadata.labels["vm.kubevirt.io/template.namespace"]="kubevirt"' | \
     oc apply -n $namespace -f -
     
     # start vm
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -110,10 +110,9 @@ EOF
     fi
 fi
 echo "Deploying CDI"
-#export CDI_VERSION=$(curl -s https://api.github.com/repos/kubevirt/containerized-data-importer/releases | \
-#            jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
+export CDI_VERSION=$(curl -s https://api.github.com/repos/kubevirt/containerized-data-importer/releases | \
+            jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
 
-export CDI_VERSION="v1.29.0"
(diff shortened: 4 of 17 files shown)


dominikholler:
@yaacov Might this change break UI?

yaacov:
> @yaacov Might this change break UI?

Yes, thanks for the heads up !
the UI will not find a boot source and complain ... we need to address this in the UI before we release this new templates.

ksimon1:
> @ksimon1 hi,
> 
> why do you move the:
> 
> ```
> source:
>           pvc:
>             name: ${SRC_PVC_NAME}
>             namespace: ${SRC_PVC_NAMESPACE}
> ```
> 
> do use:
> 
> ```
> dataSource:
>             kind: PersistentVolumeClaim
>             namespace: ${SRC_PVC_NAMESPACE}
>             name: ${SRC_PVC_NAME}
> ```
> 
> can `storage` use `source` like the `pvc` api ?
> 
> The UI will not crash if the `source` api is used instead of `dataSource`

According to https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/apis/core/v1beta1/types.go#L69, the storage doesn't have `source` attribute (@awels please keep me honest here). I moved the `source` to `storage`, because you can't have both `source` and `storage` defined, otherwise you hit https://github.com/kubevirt/containerized-data-importer/blob/cab586ab1a45f2bd1d8c81e55413f2b63b011628/pkg/apiserver/webhooks/datavolume-validate.go#L97

yaacov:
> According to https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/apis/core/v1beta1/types.go#L69, the storage doesn't have `source` attribute (@awels please keep me honest here). I moved the `source` to `storage`, because you can't have both `source.pvc` and `storage` defined, otherwise you hit https://github.com/kubevirt/containerized-data-importer/blob/cab586ab1a45f2bd1d8c81e55413f2b63b011628/pkg/apiserver/webhooks/datavolume-validate.go#L97

[datavolume-validate.go#L97](https://github.com/kubevirt/containerized-data-importer/blob/cab586ab1a45f2bd1d8c81e55413f2b63b011628/pkg/apiserver/webhooks/datavolume-validate.go#L97)
does not check for  `spec.Source` it check for  `spec.PVC` conflicting with `spec.Storage` 
that makes sense to me

can we have  `spec.Source` with `spec.Storage`  ?
http://kubevirt.io/api-reference/master/definitions.html#_v1alpha1_datavolumespec

The problem with using the intenal `spec.Storage.dataSource` instead of `spec.source` is that `dataSource` only support [LocalObjectReference](http://kubevirt.io/api-reference/master/definitions.html#_k8s_io_api_core_v1_typedlocalobjectreference)  while `source` [supports](http://kubevirt.io/api-reference/master/definitions.html#_v1alpha1_datavolumesource) more options

Since the UI is expected to manipulate this field from the UI point of view it is better to only change the `spec.source` field then remove the `spec.Storage.dataSource` and introduce a new field that is missing in other templates.

@awels what am I missing here ?

brybacki:
> > @ksimon1 hi,
> > why do you move the:
> > ```
> > source:
> >           pvc:
> >             name: ${SRC_PVC_NAME}
> >             namespace: ${SRC_PVC_NAMESPACE}
> > ```
> > 
> >   
> > do use:
> > ```
> > dataSource:
> >             kind: PersistentVolumeClaim
> >             namespace: ${SRC_PVC_NAMESPACE}
> >             name: ${SRC_PVC_NAME}
> > ```

> >   
> > can `storage` use `source` like the `pvc` api ?
> > The UI will not crash if the `source` api is used instead of `dataSource`
> 
> According to https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/apis/core/v1beta1/types.go#L69, the storage doesn't have `source` attribute (@awels please keep me honest here). I moved the `source` to `storage`, because you can't have both `source.pvc` and `storage` defined, otherwise you hit https://github.com/kubevirt/containerized-data-importer/blob/cab586ab1a45f2bd1d8c81e55413f2b63b011628/pkg/apiserver/webhooks/datavolume-validate.go#L97

But the `dv.spec.storage` and `dv.spec.source` are two different - indepenedent fields. 

The only thing that changes is that instead of `dv.spec.pvc` you can use `dv.spec.storage`. Right now `storage` has all the same fields as pvc. The only difference is the new automation for missing fields. In theory we did not need to introduce new field, but as the new logic changes semantics we need new field to be fully backwards compatible. To not surprise users .

 Other DataVolume fields are not changed. So `dv.spec.source` is still a `dv.spec.source`.

The validator only checks if there is no  `dv.spec.pvc` together with `dv.spec.storage`.

templates/centos6.tpl.yaml
```diff
@@ -66,16 +66,16 @@ objects:
       metadata:
         name: ${NAME}
       spec:
-        pvc:
+        storage:
+          dataSource:
```
brybacki, line 70: This should be a source at DataVolume level as it was previously. To correctly use CDI import logic. 

By defining dataSource here, you are using the `pvc.dataSource` which is a kubernetes extension (https://docs.openshift.com/container-platform/4.6/storage/container_storage_interface/persistent-storage-csi-cloning.html) that only supports some storage providers.

ksimon1: you are right. I was mistify by one of our script, which modified wrong value. That was why it showed the error messages

templates/centos6.tpl.yaml
```diff
@@ -66,7 +66,7 @@ objects:
       metadata:
         name: ${NAME}
       spec:
-        pvc:
+        storage:
           accessModes:
             - ReadWriteMany
```
aglitke, line 71: One principle benefit of using 'storage' is that you no longer need to specify the accessModes parameter.  CDI will look up the storage class and select the best access mode based on what provisioner is being used.  Would it be possible to remove accessModes from each of these templates?

ksimon1: Sure I can remove it

ksimon1:
/retest
@yaacov, @aglitke, @brybacki can you plese review?

yaacov:
lgtm

awels:
Note the source is for cloning operations, and dataSource is for the new data import cron.

PR #382, opened 2021-11-12
add patch which sets default accessMode and volumeMode

ksimon1:
add patch which sets default accessMode and volumeMode
bump kubernetes version

This PR updates refreshes container image automation. Previously
automation failed, due to missing informations about acessMode and
volumeMode. With this change script always sets acessMode and volumeMode
to local storageProfile.
/cc @akrejcir 

**Release note**:
```
NONE
```

### DIFF
--- a/automation/test-linux.sh
+++ b/automation/test-linux.sh
@@ -23,6 +23,7 @@ elif [[ $TARGET =~ refresh-image.* ]]; then
   image_url="docker://registry:5000/disk"
   # Inform CDI the local registry is insecure
   oc patch configmap cdi-insecure-registries -n cdi --type merge -p '{"data":{"mykey": "registry:5000"}}'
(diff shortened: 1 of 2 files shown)


automation/test-linux.sh
```diff
@@ -23,6 +23,7 @@ elif [[ $TARGET =~ refresh-image.* ]]; then
   image_url="docker://registry:5000/disk"
   # Inform CDI the local registry is insecure
   oc patch configmap cdi-insecure-registries -n cdi --type merge -p '{"data":{"mykey": "registry:5000"}}'
+  oc storageProfile local --type merge -p '{"spec": {"claimPropertySets":[{"accessModes": ["ReadWriteOnce"], "volumeMode": "Filesystem"}]}}'
```
akrejcir, line 26: This line is missing the `patch` command:
```bash
oc patch storageprofile ...
```

PR #384, opened 2021-11-22
Add a RHEL8 template targeting SAP HANA as workload

rmohr:

This adds a tiny profile for SAP hana machines based on RHEL8.

Fixes #

**Release note**:
```release-note
Add a tech-preview SAP HANA template for RHEL8
```


rmohr:
/cc @ksimon1 
/cc @dominikholler 

would be glad to get a review. I am not sure how much sense it will make to add editable fields, since we will need the template as openshift template. Happy to get some feedback on that.

templates/saphana.tpl.yaml
```diff
@@ -0,0 +1,188 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
+  annotations:
+    openshift.io/display-name: "Red Hat Enterprise Linux 8.4 Beta VM for SAP HANA workloads"
+    description: >-
+      Template for Red Hat Enterprise Linux 8.4 for SAP HANA workloads.
```
ksimon1, line 8: Is this template going to be supported only for RHEL 8.4 and newer? 
Because currently this template will be generated containing all rhel 8 versions.
e.g. see annotations
```
    name.os.template.kubevirt.io/rhel8.0: Red Hat Enterprise Linux 8.0 or higher
    name.os.template.kubevirt.io/rhel8.1: Red Hat Enterprise Linux 8.0 or higher
    name.os.template.kubevirt.io/rhel8.2: Red Hat Enterprise Linux 8.0 or higher
    name.os.template.kubevirt.io/rhel8.3: Red Hat Enterprise Linux 8.0 or higher
    name.os.template.kubevirt.io/rhel8.4: Red Hat Enterprise Linux 8.0 or higher
    name.os.template.kubevirt.io/rhel8.5: Red Hat Enterprise Linux 8.0 or higher
```
and labels 
```
    os.template.kubevirt.io/rhel8.0: "true"
    os.template.kubevirt.io/rhel8.1: "true"
    os.template.kubevirt.io/rhel8.2: "true"
    os.template.kubevirt.io/rhel8.3: "true"
    os.template.kubevirt.io/rhel8.4: "true"
    os.template.kubevirt.io/rhel8.5: "true"
```

templates/saphana.tpl.yaml
```diff
@@ -0,0 +1,188 @@
...
+            vm.kubevirt.io/flavor: "{{ item.flavor }}"
+          labels:
+            kubevirt.io/domain: ${NAME}
+            kubevirt.io/size: {{ item.flavor }}
+        spec:
+          hostname: ${NAME}
+          domain:
+            ioThreadsPolicy: auto
+            cpu:
+              cores: {{ '${{CPU_CORES}}' }}
```
ksimon1, line 72: shouldn't this have syntax `${CPU_CORES}`?
Same for `threads` and `sockets`

templates/saphana.tpl.yaml
```diff
@@ -0,0 +1,188 @@
...
+                networkName: ${SRIOV_NETWORK_NAME2}
+              name: sriov-net2
+            - multus:
+                networkName: ${SRIOV_NETWORK_NAME3}
+              name: sriov-net3
+          nodeSelector:
+            kubernetes.io/hostname: ${TARGET_NODE_NAME}
+parameters:
+  - description: Name for the new VM
+    name: NAME
```
ksimon1, line 148: Do we want to have some default name?

generate-templates.yaml
```diff
@@ -37,11 +37,24 @@
...
   - name: Load RHEL 8 versions
     set_fact:
       rhel8_labels: "{{ lookup('osinfo', 'distro=rhel') |map(attribute='short_id') |select('match', '^rhel8\\.') |list |sort }}"
 
+  - name: Generate RHEL 8 SAP HANA template
+    template:
+      src: saphana.tpl.yaml
+      dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}.yaml"
+    with_items:
+      - {flavor: tiny,  workload: saphana, memsize: "24Gi",   cpus: 4, iothreads: True,  emulatorthread: True,  tablet: False, default: False}
```
ksimon1, line 49: You don't have to specify `memsize` and `cpus` here, since you specify them as template parameters

ksimon1 (changes requested):
I am missing e2e test for this new template

templates/saphana.tpl.yaml
rmohr: No, it should be filled in as numeric value and not as string: https://docs.openshift.com/container-platform/4.7/openshift_images/using-templates.html#templates-writing-parameters_using-templates

templates/saphana.tpl.yaml
rmohr: I think specifying a name is a requierment. But now that you bring it up, probably not specifying a default for the target node name may make sense. What do you think?

templates/saphana.tpl.yaml
rmohr: I think it has to be 8.4 right now. Can I just directly set the label instead of the generate?

generate-templates.yaml
rmohr: Done.

templates/saphana.tpl.yaml
ksimon1: If we want to pinpoint it to only 8.4, then it is fine to set annotation and label with exact version

templates/saphana.tpl.yaml
```diff
@@ -0,0 +1,183 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
+  annotations:
+    openshift.io/display-name: "Red Hat Enterprise Linux 8.4 Beta VM for SAP HANA workloads"
```
ksimon1, line 6: Does the word `Beta` here has meaning rhel 8.4 beta or beta sap template?

yaacov:
> @yaacov does this new template needs any UI change?

yes, we are prioritizing the changes for 4.10
(p.s. until changes are done in the UI (we plan it to be done for 4.10), we should not document that this template works with the UI)

cc:// @tnisan

templates/saphana.tpl.yaml
rmohr: It is just a copy-pase error. :+1:

rmohr: Fixed.

templates/saphana.tpl.yaml
rmohr: Done.

templates/saphana.tpl.yaml
rmohr: Also removed the default value from TARGET_NODE_NAME.

rmohr:
This template probably need https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1621 merged, so that the NUMA feature gate is enabled for kubevirt. @ksimon1 what HCO builds is common-templates using?

ksimon1:
> This template probably need [kubevirt/hyperconverged-cluster-operator#1621](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1621) merged, so that the NUMA feature gate is enabled for kubevirt. @ksimon1 what HCO builds is common-templates using?

If we should wait, before that PR is merged, then it is not a problem, because anyway we have to wait with release of new templates for new features from UI. And HCO will not use this new template until we do a new ssp operator release (which will contain new release of templates).

dominikholler (changes requested):
@rmohr Can you please add validations for the parameters.

templates/saphana.tpl.yaml
```diff
@@ -0,0 +1,179 @@
...
+            machine:
+              type: ""
+            resources:
+              requests:
+                memory: ${MEMORY_OVERHEAD}
+            memory:
+              guest: ${MEMORY}
+              hugepages:
+                pageSize: ${HUGEPAGES_PAGE_SIZE}
+          terminationGracePeriodSeconds: 0
```
ksimon1, line 111: According to @rnetser terminationGracePeriodSeconds can't be zero

rmohr: Thanks @rnetser that is a beginner mistake :)

rmohr: Fixed.

automation/unit-tests.sh
```diff
@@ -6,26 +6,30 @@ make generate
 templates=`ls dist/templates/*`
 namespace="kubevirt"
 
-oc apply -f - <<EOF
```
ksimon1, line 9: can you please keep creating of namespace here? Unit tests lane is failing due to this change

rmohr: Done. :+1:

rmohr:
@ksimon1 seems like the validation tries to literally use my `${MEMORY_OVERHEAD}` field as memory input:

```
ERROR:root:could not convert string to float: '${MEMORY_OVERHEAD}'
```

Trying to make it resolve the variable and taking the default.

templates/saphana.tpl.yaml
```diff
@@ -0,0 +1,182 @@
...
+              name: sriov-net3
+          nodeSelector:
+            kubernetes.io/hostname: ${TARGET_NODE_NAME}
+parameters:
+  - description: Name for the new VM
+    name: NAME
+    required: true
+  - description: Amount of memory
+    name: MEMORY
+    value: 24Gi
```
ksimon1, line 148: is ok when user insert lower memory or fewer e.g. cpu cores/threads/sockets then value in defaults?
If not you can use validation rules https://github.com/kubevirt/common-templates/blob/master/templates/VALIDATION.md

rmohr: I want to keep validation out for now on this template. But good to know. Thanks.

rmohr:
@ksimon1 unit tests were finally passing: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kubevirt_common-templates/384/pull-ci-kubevirt-common-templates-master-e2e-azure-common-templates-unit-tests/1466323221786660864

PTAL :)

templates/saphana.tpl.yaml
rmohr: Memory validation added. :+1:

ksimon1:
/lgtm
@dominikholler can you take a look please

dominikholler:
Would have the 'workload.template.kubevirt.io/server: "true"' any semantics?
Are the "template.kubevirt.io/provider*" annotations added in a later stage?

ksimon1:
> Are the "template.kubevirt.io/provider*" annotations added in a later stage?
This annotation is added later.

rnetser:
LGTM

rmohr:
:+1:

PR #385, opened 2021-11-29
Change DataVolume source from PVC to DataSource

akrejcir:
This indirection allows easier update of PVCs, because the DataSource can be updated when a new PVC is imported.

**Release note**:
```release-note
Changed DataVolume source from PVC to DataSource.
```

### DIFF
--- a/automation/test-linux.sh
+++ b/automation/test-linux.sh
@@ -28,11 +28,13 @@ else
   image_url="docker://quay.io/kubevirt/common-templates:${TARGET}"
 fi
 
+dv_name="${TARGET}-datavolume-original"
+
 oc apply -n $namespace -f - <<EOF
 apiVersion: cdi.kubevirt.io/v1beta1
 kind: DataVolume
 metadata:
-  name: ${TARGET}-datavolume-original
+  name: ${dv_name}
 spec:
   source:
     registry:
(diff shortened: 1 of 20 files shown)


automation/test-linux.sh
```diff
@@ -49,7 +51,19 @@ EOF
 timeout=600
 sample=10
 
-oc wait --for=condition=Ready --timeout=${timeout}s dv/${TARGET}-datavolume-original -n $namespace
+oc wait --for=condition=Ready --timeout=${timeout}s dv/${dv_name} -n $namespace
+
+oc apply -n $namespace -f - <<EOF
+apiVersion: cdi.kubevirt.io/v1beta1
```
ksimon1, line 57: isn't DataSource automatically created when you create VM from template?

akrejcir: I think not, but I haven't tried it.
Where is the code or component that would create the DataSoruce ?

akrejcir: And it would not make sense to create the `DataSource` when the **VM** is created. The `DataSource` has to exist and point to the latest golden image PVC when VM is created.

akrejcir:
The `DataSoruces` will be created by the SSP operator when creating the templates.

ksimon1:
/lgtm
@tnisan can someone from UI team review this PR and put lgtm label?

PR #389, opened 2021-12-09
remove generation of sap hana template

ksimon1:
remove generation of sap hana template

We don't want to ship sap hana template in current release

**Release note**:
```
Remove sap hana template
```

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -40,20 +40,20 @@
   - name: Load RHEL 8 versions
     set_fact:
       rhel8_labels: "{{ lookup('osinfo', 'distro=rhel') |map(attribute='short_id') |select('match', '^rhel8\\.') |list |sort }}"
-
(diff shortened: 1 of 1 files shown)


generate-templates.yaml
```diff
@@ -40,20 +40,20 @@
...
-      - {flavor: tiny,  workload: saphana, default: False}
-    vars:
-      os: rhel8
-      icon: rhel
-      majorrelease: rhel8
-      oslabels: "{{ rhel8_labels }}"
-      osinfoname: "{{ oslabels[0] }}"
-      cloudusername: cloud-user
+#
+#  - name: Generate RHEL 8 SAP HANA template
```
rmohr, line 44: How about just adding something like

```
when: 0 > 1
```

then people can enable it easily if needed.

akrejcir: I will add it in another PR. Karel is on pto now, and I don't have permissions to edit this PR.

PR #392, opened 2021-12-13
Add a condition to saphana ansible task to disable it

akrejcir:
Added 'when: 0 > 1' to the ansible task.

**Release note**:
```release-note
None
```

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -40,20 +40,22 @@
   - name: Load RHEL 8 versions
     set_fact:
       rhel8_labels: "{{ lookup('osinfo', 'distro=rhel') |map(attribute='short_id') |select('match', '^rhel8\\.') |list |sort }}"
-#
-#  - name: Generate RHEL 8 SAP HANA template
-#    template:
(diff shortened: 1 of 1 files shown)


generate-templates.yaml
```diff
@@ -40,20 +40,22 @@
...
-#      os: rhel8
-#      icon: rhel
-#      majorrelease: rhel8
-#      oslabels: "{{ rhel8_labels }}"
-#      osinfoname: "{{ oslabels[0] }}"
-#      cloudusername: cloud-user
+
+  # The SAP HANA template will not be generated
+  - name: Generate RHEL 8 SAP HANA template
+    when: 0 > 1 # This template is disabled
```
borod108, line 46: why don't we delete this and then pull it back from git? Is this for some kind of visibility?

borod108:
I added a small question, but it is not critical.
/lgtm

generate-templates.yaml
akrejcir: Yes. See comment on previous PR: https://github.com/kubevirt/common-templates/pull/389#discussion_r767539174

PR #397, opened 2021-12-30
Change recommended disk image url 32 -> 34

RoniKishner:
Change the image download URL recommended on fedora templates from 32 -> 34 since the templates are created with Fedora 34+

**Release note**:
```
NONE
```

### DIFF
--- a/templates/fedora.tpl.yaml
+++ b/templates/fedora.tpl.yaml
@@ -8,7 +8,7 @@ metadata:
       Template for {{ lookup('osinfo', oslabels[0]).name }} VM or newer.
       A PVC with the Fedora disk image must be available.
       Recommended disk image:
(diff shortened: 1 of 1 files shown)


templates/fedora.tpl.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
       Template for {{ lookup('osinfo', oslabels[0]).name }} VM or newer.
       A PVC with the Fedora disk image must be available.
       Recommended disk image:
-      https://download.fedoraproject.org/pub/fedora/linux/releases/32/Cloud/x86_64/images/Fedora-Cloud-Base-32-1.6.x86_64.qcow2
+      https://download.fedoraproject.org/pub/fedora/linux/releases/34/Cloud/x86_64/images/Fedora-Cloud-Base-34-1.2.x86_64.qcow2
```
ksimon1, line 11: Currently the fedora templates supports versions 33, 34, 35. I would vote for using version 33 or 35 image.

RoniKishner: Changed to 35

PR #398, opened 2022-01-03
Rename parameters to DATA_SOURCE_NAME and DATA_SOURCE_NAMESPACE

akrejcir:
Rename from SRC_PVC_NAME to DATA_SOURCE_NAME,
and from SRC_PVC_NAMESPACE to DATA_SOURCE_NAMESPACE.

**Release note**:
```release-note
Renamed parameters in templates.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ $ oc process --local -f dist/templates/windows10-desktop-medium.yaml
 $ oc process --local -f dist/templates/windows10-desktop-medium.yaml  --parameters
 NAME                    DESCRIPTION                       GENERATOR           VALUE
 NAME                    VM name                           expression          windows-[a-z0-9]{6}
-SRC_PVC_NAME        Name of the DataSource to clone                       win10
-SRC_PVC_NAMESPACE   Namespace of the DataSource                           kubevirt-os-images
+DATA_SOURCE_NAME        Name of the DataSource to clone                       win10
+DATA_SOURCE_NAMESPACE   Namespace of the DataSource                           kubevirt-os-images
 
 $ oc process --local -f dist/templates/windows10-desktop-medium.yaml | kubectl apply -f -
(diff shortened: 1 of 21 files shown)


ksimon1:
@akrejcir UI is ready for this change?

akrejcir:
They are working on it right now.

akrejcir:
Waiting for UI
/hold

dominikholler:
@akrejcir Can you please share the motivation for this change in the commit message?

dominikholler:
@akrejcir Can you please share the motivation for this change in the commit message?
@yaacov Is UI ready to get this update?

yaacov:
> @yaacov Is UI ready to get this update?

No, unfortunately the current UI assume the PVC source is in a parameter called  `SRC_PVC_NAME/NAMESPCE`

p.s. Once we move to using parameters "as is" this will no longer be a problem, but in 4.10 the parameter name is hardcoded in the UI.

dominikholler:
> No, unfortunately the current UI assume the PVC source is in a parameter called SRC_PVC_NAME/NAMESPCE
> p.s. Once we move to using parameters "as is" this will no longer be a problem, but in 4.10 the parameter name is hardcoded in the UI.

@yaacov Do you plan to modify the name of the parameter in 4.10 ?

yaacov:
> @yaacov Do you plan to modify the name of the parameter in 4.10 ?

We didn't know about the name change, so we didn't make any plans

Note I:
if we do support it, we will have to support both conventions, and check for both names because we need to be backword compatible,  
Note II:
It will be nice if we wait with the name change to 4.11, where we plan to remove the practice of giving meaning to template parameters in the UI (CNV-14778)

cc:// @tnisan

ksimon1:
/cherrypick release-v0.19
Backporting forgotten PR

PR #399, opened 2022-01-06
Centos stream

ksimon1:
add centos stream 8 and 9 templates 
update osinfo

delete centos 8 template because centos 8 is EOL

**Release note**:
```
Drop centos 8 template because centos 8 is EOL
Add centos stream 8 and 9
```

### DIFF
--- a/automation/test-linux.sh
+++ b/automation/test-linux.sh
@@ -85,7 +85,7 @@ if [[ $TARGET =~ opensuse.* ]]; then
   workloads=("server")
 fi
 
-if [[ $TARGET =~ centos7.* ]] || [[ $TARGET =~ centos8.* ]]; then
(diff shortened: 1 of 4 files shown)


ksimon1:
this PR https://github.com/openshift/release/pull/24986 has to be merged as well

dominikholler (approved):
@rnetser @yaacov Please note that this change changes more than just the "openshift.io/display-name" , this is a clean update which changes also the annotation "name.os.template.kubevirt.io/centos8" to "name.os.template.kubevirt.io/centos-stream8", has the label "os.template.kubevirt.io/centos-stream8" instead of "os.template.kubevirt.io/centos8", and uses "stream" inside "vm.kubevirt.io/template" and "vm.kubevirt.io/os". Please let us know if this creates any trouble.

/lgtm

PR #401, opened 2022-01-13
Adjust hana template parameter details

rmohr:

This PR improves the following aspects of the hana template:
 * Use the `registry.redhat.io` registry instead of `registry.access.redhat.com`
 * Make SRIOV parameters required, since they always have to be adjusted to the environment
 * Give users the freedom to set the node selectors freely
 * Add a blanko toleration which helps scheduling the VM to exclusively dedicated nodes

Fixes #

**Release note**:

```release-note
Use the correct RHEL containerdisk URL, make SRIOV resource names required and make node selection more flexible
```

### DIFF
--- a/automation/unit-tests.sh
+++ b/automation/unit-tests.sh
@@ -24,7 +24,7 @@ done
 echo "Processing all templates to find syntax issues"
 for template in $templates; do
   if [[ $template =~ .*saphana.* ]]; then
-    oc process -f "$template" NAME=test TARGET_NODE_NAME=mynode || exit 1
+    oc process -f "$template" NAME=test WORKLOAD_NODE_LABEL_VALUE="somevalue" SRIOV_NETWORK_NAME1="default/sriov1"  SRIOV_NETWORK_NAME2="default/sriov1"  SRIOV_NETWORK_NAME3="default/sriov1"|| exit 1
   else
     oc process -f "$template" NAME=test SRC_PVC_NAME=test || exit 1
   fi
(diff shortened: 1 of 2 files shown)


automation/unit-tests.sh
```diff
@@ -24,7 +24,7 @@ done
 echo "Processing all templates to find syntax issues"
 for template in $templates; do
   if [[ $template =~ .*saphana.* ]]; then
-    oc process -f "$template" NAME=test TARGET_NODE_NAME=mynode || exit 1
+    oc process -f "$template" NAME=test TARGET_NODE_SELECTOR='{"mynode/lala": "blub"}' SRIOV_NETWORK_NAME1="default/sriov1"  SRIOV_NETWORK_NAME2="default/sriov1"  SRIOV_NETWORK_NAME3="default/sriov1"|| exit 1
```
rmohr, line 27: @rnetser I think one caveat of this approach is that when we do not pre-define certain aspects of the template, that people have to provide here a valid json expression.

Another option would be solely allow setting a key, or just a value of a label. I am not sure if the UI can handle well what I am doing here.

ksimon1:
/lgtm
@rnetser can you please review

automation/unit-tests.sh
rnetser: I think that we should provide the simplest way for users to set the label.
Is it a legitimate requirement to ask to set the label in a specific format, for example:
` node-role.kubernetes.io/sap: "" ` and then only parametrize the suffix of the template key (`node-role.kubernetes.io/${NODE_LABEL}: "" `)?

rmohr: That sounds good. One more thing came to my mind. We will also have to give people the chance to reserve the whole node for hana. That means that people have to taint nodes to avoid that other workloads are going there. I will add a configurable toleration to the template too.

rnetser: Ack

rmohr:
@rnetser @ksimon1 can you have another look?

rnetser:
LGTM

PR #404, opened 2022-01-19
Remove virtio model reference on sriov

rmohr:

The kubevirt validation allows passing through models on sriov devices
which it should not, and they have no effect. Removing them.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/rhel8.saphana.tpl.yaml
+++ b/templates/rhel8.saphana.tpl.yaml
@@ -107,13 +107,10 @@ objects:
                   model: virtio
                 - name: sriov-net1
                   sriov: {}
-                  model: virtio
                 - name: sriov-net2
                   sriov: {}
-                  model: virtio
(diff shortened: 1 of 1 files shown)


rmohr:
@dominikholler as you suspected, the model is ignored.

dominikholler:
@rmohr thanks for follow up. The change looks good to me. `networkInterfaceMultiqueue` might be still ignored, but if we keep it, it might be a good hint for users who wants to test with virtio virtual NICs.
@EdDev  are you fine with this code change?

EdDev:
> EdDev are you fine with this code change?

Looks right to me.
We should probably add warning events in the validation, but unfortunately we cannot block it there (due to backward compatibility).
Is it worth adding an event with a warning @ kubevirt/kubevirt?

EdDev (approved):

PR #406, opened 2022-02-10
Add annotations for containerdisks and import URLs

0xFelix:
**What this PR does / why we need it**:

This adds the annotations `template.kubevirt.io/containerdisks` and
`template.kubevirt.io/images` to the common templates where appropriate.

The annotations can be used to automatically download guest images or to
display instructions for the user.

Both annotations contain lists where each element is separated by a new
line.

Image URLs are automatically obtained from the osinfo-db. Containerdisks are
static for now. (until containerdisks are added to libosinfo)

**Which issue(s) this PR fixes**

https://bugzilla.redhat.com/show_bug.cgi?id=2031155
https://bugzilla.redhat.com/show_bug.cgi?id=2038832
https://bugzilla.redhat.com/show_bug.cgi?id=2031857

**Special notes for your reviewer**:

Merge after #412 and #413, rebase first

**Release note**:

```release-note
Add template.kubevirt.io/containerdisks and template.kubevirt.io/images annotations
```

PR #407, opened 2022-02-10
Update OpenSUSE to 15.3

0xFelix:

Update OpenSUSE to 15.3

Merge after #406 

**Release note**:

```release-note
Update OpenSUSE to 15.3
```

### DIFF
--- a/templates/opensuse.tpl.yaml
+++ b/templates/opensuse.tpl.yaml
@@ -7,8 +7,6 @@ metadata:
     description: >-
       Template for OpenSUSE Leap 15.0 VM.
       A PVC with the OpenSUSE disk image must be available.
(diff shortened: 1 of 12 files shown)


templates/opensuse.tpl.yaml
```diff
@@ -7,8 +7,6 @@ metadata:
     description: >-
       Template for OpenSUSE Leap 15.0 VM.
```
dominikholler, line 8: `Template for OpenSUSE Leap 15 VM.`

dominikholler (changes requested):
Why is  OpenSUSE not updated automatically like fedoera?
cc: @vasiliy-ul

0xFelix:
See #413

PR #408, opened 2022-02-10
Add Ubuntu server workload

0xFelix:

Add Ubuntu server workload

Only server workloads have cloud images, this allows to display import urls after #406 is merged. 

**Release note**:

```release-note
Add Ubuntu server workload
```


0xFelix:
See #412

PR #409, opened 2022-02-11
Update templates/LABELS.md

0xFelix:

This file is an additional maintenance burden with no real use.

**Release note**:

```release-note
None
```

### DIFF
--- a/templates/LABELS.md
+++ b/templates/LABELS.md
@@ -1,16 +1,7 @@
 # Available labels
 
-This file describes the labels we will be using to select
+This file exemplarily describes the labels we will be using to select
 the workload a template is supposed to be used for.
-
-Some of them are not going to be used immediately and serve
-as examples.
-
-## Deprecated templates
-
-When a template is deprecated, the following annotation is applied to it: `template.kubevirt.io/deprecated: "true"`.
-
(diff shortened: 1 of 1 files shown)


templates/LABELS.md
```diff
@@ -1,97 +0,0 @@
-# Available labels
-
-This file describes the labels we will be using to select
-the workload a template is supposed to be used for.
-
-Some of them are not going to be used immediately and serve
-as examples.
-
-## Deprecated templates
```
ksimon1, line 9: delete only this paragraph, not whole file

yaacov:
> @yaacov Could you have a look at this? Is this file still needed by UI?

It's just the documentation right ?

if it's not just documentation:
the UI use `deprecated` label, to not show it to users
the UI use the `os.template.kubevirt.io`, to guess the os name
workload and flavor are used to desplay nice information about the template

we need this labels in 4.10
we can deprecate them in 4.11, are the getting deprecated ?

0xFelix:
@yaacov Yes, it is only documentation. Sorry for the misunderstanding. It is tedious to update the versions in `LABELS.md` manually. The labels are definitely going to stay.

I updated the PR so LABELS.md is kept and changed the tedious parts instead.

PR #410, opened 2022-02-11
Mark CentOS 6 template as deprecated

0xFelix:

CentOS 6 is EOL and therefore marked as deprecated.

The deprecated templates notice, which was removed in 4e48e4e6, is
restored.

**Release note**:

```release-note
Mark CentOS 6 templates as deprecated
```

### DIFF
--- a/templates/LABELS.md
+++ b/templates/LABELS.md
@@ -3,6 +3,12 @@
 This file exemplarily describes the labels we will be using to select
 the workload a template is supposed to be used for.
 
+## Deprecated templates
+
+When a template is deprecated, the following annotation is applied to it: `template.kubevirt.io/deprecated: "true"`.
+
+The template itself is not removed for backward compatibility reasons.
+
 ## Operating systems
 
 The operating system labels must match the [libosinfo
--- a/templates/centos6.tpl.yaml
+++ b/templates/centos6.tpl.yaml
@@ -4,6 +4,7 @@ kind: Template
 metadata:
   name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
   annotations:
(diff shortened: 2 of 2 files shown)


dominikholler:
Are you aware of any dependencies to this template @lersek @rwmjones ?

rwmjones:
I don't understand what the question means.

rwmjones:
Nothing I know of, but I don't know much about kubevirt so I may not be the best person to ask.

lersek:
same response from me...

dominikholler:
thanks for having a look @lersek @rwmjones!
I want to ensure, that the v2v/konveyor/mtv tools does not depend on the centos 6 template.

@mansam are you aware of dependencies in konveyor to the centos 6 template.

ksimon1:
@mansam can you please confirm if there are some dependencies to centos 6 template in konveyor?

mansam:
Hi, apologies for not seeing this sooner. We don't have any hard dependencies on the centos6 template; Forklift will use it if it's available, but not having it won't break us.

0xFelix:
@yaacov @mansam Does the UI look for the `template.kubevirt.io/deprecated: "true"` label? Does the UI hide templates when the label is found? Maybe we could add the label to the CentOS 6 template instead of deleting it. This way the UI would not encourage users to create new VMs with obsolete OSes while Forklift could still use the template to migrate legacy VMs.

0xFelix:
@dominikholler: @yaacov confirmed that deprecated templates are hidden by the ui. I changed the PR so CentOS 6 is marked as deprecated instead of removing it.

yaacov:
LGTM

PR #412, opened 2022-02-14
Improve Ubuntu template

0xFelix:

- Add server workload to template and make it the default
- Remove unnecessary tiny flavor
- Automatically load active Ubuntu versions

**Release note**:

```release-note
Added Ubuntu server workload as new default
Removed Ubuntu tiny flavor
Include support for latest and active Ubuntu versions
```


0xFelix:
@dominikholler This PR also independent from #406 now.

PR #414, opened 2022-02-14
Add osinfo-db overrides for distributions without cloud images

0xFelix:

Add osinfo-db overrides for distributions without cloud images, so
templates with image urls can be generated. 

Remove these overrides when cloud images are added to the distributions
in upstream osinfo-db.

**Release note**:

```release-note
None
```


dominikholler:
The xml files contains too much information for us to maintain, please find a way to add only the url.

0xFelix:
Adding the whole XML was not necessary, you are right. It is sufficient to only add the URLs to the override.

dominikholler (approved):
/lgtm

PR #416, opened 2022-02-21
modify fedora template to start with secure boot support

Barakmor1:
modify fedora template that support UEFI BIOS with secure boot by default.

Fixes #

- UEFI BIOS is work with fedora35+

**Release note**:

```release-note
modify fedora template to have secure boot support
```

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -238,7 +238,7 @@
 
   - name: Load Fedora versions
     set_fact:
-      fedora_labels: "{{ lookup('osinfo', 'distro=fedora') |select('osinfo_active') |map(attribute='short_id') |select('match', '^fedora\\d+$') |list |sort }}"
+      fedora_labels: "{{ lookup('osinfo', 'distro=fedora') |select('osinfo_active') |map(attribute='short_id') |select('match', '^fedora3[5-9]|^fedora[4-9][0-9]') |list |sort }}"
 
   - name: Load Fedora image urls
     set_fact:
--- a/templates/fedora.tpl.yaml
+++ b/templates/fedora.tpl.yaml
@@ -93,6 +93,12 @@ objects:
           kubevirt.io/size: {{ item.flavor }}
       spec:
         domain:
+          features:
+            smm:
+              enabled: true
+          firmware:
+            bootloader:
+              efi: {}
 {% if item.iothreads %}
           ioThreadsPolicy: shared
 {% endif %}


dominikholler:
@ksimon1 and @Barakmor1 Do you think we have to keep the fedora 34 template? Fedora 34 might be eol during may.

Barakmor1:
> @ksimon1 and @Barakmor1 Do you think we have to keep the fedora 34 template? Fedora 34 might be eol during may.

Hey @dominikholler ,
I personally think we should keep fedora 34 for a while as fedora 35 is relatively new OS,
i think that this is out of the scope of this PR .

ksimon1:
Currently fedora template supports versions 34 and 35. I would vote for updating fedora template to only use efi. @dominikholler  has valid point that F34 is going EOL in less than 3 months and 4.11 will be released after this date. Because otherways we will have to keep both templates with efi and without with different names (e.g. fedora-server-large-efi).

Barakmor1:
> Currently fedora template supports versions 34 and 35. I would vote for updating fedora template to only use efi. @dominikholler has valid point that F34 is going EOL in less than 3 months and 4.11 will be released after this date. Because otherways we will have to keep both templates with efi and without with different names (e.g. fedora-server-large-efi).

Hey @ksimon1 ,
I tried to use the old template with efi as default but it doesn't work with fedora <= 34:
image

Barakmor1:
> Currently fedora template supports versions 34 and 35. I would vote for updating fedora template to only use efi. @dominikholler has valid point that F34 is going EOL in less than 3 months and 4.11 will be released after this date. Because otherways we will have to keep both templates with efi and without with different names (e.g. fedora-server-large-efi).

Thanks!
I will update the PR soon .

dominikholler:
@Barakmor1 does the template work with fedora 36, e.g. the qcow2 from https://download.fedoraproject.org/pub/fedora/linux/development/36/Cloud/x86_64/images/ ?

Barakmor1:
> @Barakmor1 does the template work with fedora 36, e.g. the qcow2 from https://download.fedoraproject.org/pub/fedora/linux/development/36/Cloud/x86_64/images/ ?

i will try it and update :)

dominikholler:
@Barakmor1 I wonder if this change applies only to fedora, or if other linux distributions, e.g. centos stream 9, are affected, too.
Do you think it is implementable to derive the decision if efi or bios should be used from libosinfo?

dominikholler:
> Do you think it is implementable to derive the decision if efi or bios should be used from libosinfo?

No, the libosinfo attribute seems to refer to the guest os in general, not to the cloud image.

Barakmor1:
@dominikholler 
Hey,
I updated the PR to modify the old template instead of adding new one,
also i tested it and the new template work with fedora 36.

generate-templates.yaml
```diff
@@ -238,7 +238,7 @@
 
   - name: Load Fedora versions
     set_fact:
-      fedora_labels: "{{ lookup('osinfo', 'distro=fedora') |select('osinfo_active') |map(attribute='short_id') |select('match', '^fedora\\d+$') |list |sort }}"
+      fedora_labels: "{{ lookup('osinfo', 'distro=fedora') |select('osinfo_active') |map(attribute='short_id') |select('match', '^fedora3[5-9]|^fedora[4-9][0-9]') |list |sort }}"
```
dominikholler, line 242: Probably this can be simplified again after the release of fedora36

dominikholler (approved):
/lgtm

PR #417, opened 2022-02-23
Add containerdisk annotations test

0xFelix:
**What this PR does / why we need it**:

This adds a test for the appropriate templates which ensures that the
actual templates in a live cluster contain the containerdisk
annotations.

**Special notes for your reviewer**:

Merge after #418

**Release note**:

```release-note
None
```

PR #418, opened 2022-02-24
Cleanup builder/Dockerfile

0xFelix:

Cleanup `builder/Dockerfile` and prepare it for containerdisk annotations test. (#417)

- Use quay.io/fedora/fedora:latest as base
- Remove unused dependencies / tools
- Cleanup syntax
- Download latest oc client binary
- Add latest yq binary (needed for containerdisk annotations test)
- Add comments

**Release note**:

```release-note
None
```

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -1,14 +1,20 @@
-#This docker file is used in openshift CI
-FROM quay.io/kubevirt/common-templates:fedora33CI
+# This Dockerfile is used in openshift CI
+FROM quay.io/fedora/fedora:latest
 
-RUN dnf install -y jq ansible libosinfo python3-gobject python3-openshift libosinfo intltool make git findutils qemu expect golang
-RUN chmod uga+w /etc/passwd
+# Install dependencies and tools
+RUN dnf install -y jq ansible python3-gobject python3-openshift libosinfo make git findutils expect golang
 
-#set permissions for ansible tmp folder
-RUN mkdir -p /.ansible/tmp
-RUN chmod -R 777 /.ansible
+# Allow writes to /etc/passwd so a user for ansible can be added by CI commands
+RUN chmod a+w /etc/passwd
 
-# download oc client binary
-RUN curl https://mirror2.openshift.com/pub/openshift-v4/clients/oc/4.6/linux/oc.tar.gz | tar -C . -xzf -
-RUN chmod +x oc
-RUN mv oc /usr/bin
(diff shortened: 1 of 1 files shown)


builder/Dockerfile
```diff
@@ -1,14 +1,20 @@
-#This docker file is used in openshift CI
-FROM quay.io/kubevirt/common-templates:fedora33CI
+# This Dockerfile is used in openshift CI
+FROM quay.io/fedora/fedora:latest
```
akrejcir, line 2: Can you use a fixed version instead of `latest`? Just to be sure that it is not unexpectedly updated.

akrejcir (commented):
I'm just curious, why did you separate this PR from https://github.com/kubevirt/common-templates/pull/417 ?

0xFelix:
@akrejcir: @ksimon1 told me to do so because the CI would not pick up the new `builder/Dockerfile` otherwise. The test introduced #417 depends on it and currently fails.

builder/Dockerfile
0xFelix: From what I've seen `latest` points to the latest Fedora `stable` release. Ideally, we would always use that anyway? Currently it is an additional maintenance task to update the image whenever a new Fedora `stable` is released.

akrejcir: Ok

akrejcir:
> @akrejcir: @ksimon1 told me to do so because the CI would not pick up the new `builder/Dockerfile` otherwise. The test introduced in #417 depends on it and currently fails.

All right, I didn't know that.

builder/Dockerfile
```diff
@@ -1,14 +1,20 @@
...
-# download oc client binary
-RUN curl https://mirror2.openshift.com/pub/openshift-v4/clients/oc/4.6/linux/oc.tar.gz | tar -C . -xzf -
-RUN chmod +x oc
-RUN mv oc /usr/bin
+# Create ansible tmp folder and set permissions
+RUN mkdir -p /.ansible/tmp && \
+    chmod -R 777 /.ansible
+
+# Download oc client binary
+RUN curl https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/oc/latest/linux/oc.tar.gz | tar -C /usr/local/bin -xzf - oc && \
```
ksimon1, line 15: We might go into troubles with latest oc in the future. Please use the same oc client version as we are using for openshift in e2e tests

PR #421, opened 2022-02-25
Follow redirects when downloading oc and yq

0xFelix:

Follow redirects (if there are any) when downloading oc and yq binaries
in `builder/Dockerfile`, so instead of a redirection page the actual
binaries are received.

**Release note**:

```release-note
None
```


ksimon1:
Merging to unblock ci testing

PR #441, opened 2022-05-12
bump fedora containerdisk to 36

ksimon1:
bump fedora containerdisk to 36

**Release note**:
```
bump fedora containerdisk to 36
```

### DIFF
--- a/templates/fedora.tpl.yaml
+++ b/templates/fedora.tpl.yaml
@@ -16,7 +16,7 @@ metadata:
     template.kubevirt.io/version: v1alpha1
     defaults.template.kubevirt.io/disk: rootdisk
     template.kubevirt.io/containerdisks: |
-      quay.io/containerdisks/fedora:35
+      quay.io/containerdisks/fedora:36
 {% if image_urls | length > 0 %}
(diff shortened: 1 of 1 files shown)


templates/fedora.tpl.yaml
```diff
@@ -16,7 +16,7 @@ metadata:
     template.kubevirt.io/version: v1alpha1
     defaults.template.kubevirt.io/disk: rootdisk
     template.kubevirt.io/containerdisks: |
-      quay.io/containerdisks/fedora:35
+      quay.io/containerdisks/fedora:36
```
ksimon1, line 19: @dominikholler is it ok to keep only 36, or do we want to keep both?

dominikholler: Let's go with only 36 for  now, to be sure the version 36 is shown in UI.

@upalatucci if there would be multiple containerdisks, which one would be shown in the UI?

dominikholler (approved):
/lgtm
fredora 36 reached in https://quay.io/repository/containerdisks/fedora?tab=tags

PR #486, opened 2022-10-20
Windows 2022 template is missing vTPM and UEFI

Barakmor1:
Signed-off-by: bmordeha <bmodeha@redhat.com>

Windows 2022 template is missing vTPM and UEFI
Fixes #

**Release note**:

```release-note
add vTPM and UEFI to Windows 2022 template
```

### DIFF
--- a/templates/windows2k22.tpl.yaml
+++ b/templates/windows2k22.tpl.yaml
@@ -128,6 +128,7 @@ objects:
           features:
             acpi: {}
             apic: {}
+            smm: {}
             hyperv:
               relaxed: {}
               vapic: {}
@@ -143,6 +144,10 @@ objects:
               ipi: {}
               runtime: {}
               reset: {}
+          firmware:
+            bootloader:
+              efi:
+                secureBoot: true
           devices:
 {% if item.multiqueue %}
             networkInterfaceMultiqueue: True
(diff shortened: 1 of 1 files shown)


Barakmor1:
@ksimon1  
Hey,
I need to update windows2022 image to include UEFI to make it work,
but the old manifest doesn't work with the new image so i think that we should
switch the images as fast as possible.
Maybe it would be better if we will approve this PR before i'll update the images so
this PR will get merged faster.
/test e2e-azure-common-templates-rhel6
/test e2e-azure-common-templates-windows2019

dominikholler:
@Barakmor1  From my point of view it would be helpful, of the commit message would explain the "missing". In which way Win2k22 guests would benefit from using uefi and vTPM, and what would be the drawbacks relative to use BIOS?

Barakmor1:
> @Barakmor1 From my point of view it would be helpful, of the commit message would explain the "missing". In which way Win2k22 guests would benefit from using uefi and vTPM, and what would be the drawbacks relative to use BIOS?

Thanks for the comment.
I updated the commit massage.

dominikholler:
> > @Barakmor1 From my point of view it would be helpful, of the commit message would explain the "missing". In which way Win2k22 guests would benefit from using uefi and vTPM, and what would be the drawbacks relative to use BIOS?
> 
> Thanks for the comment. I updated the commit massage.

Thanks for improving the commit message!

PR #493, opened 2022-11-17
improve RHEL9 display name

dankenigsberg:
"Red Hat Enterprise Linux 9 VM" is visually cleaner and more precise display name now that RHEL 9.1 is available as is the default.

```release-note
NONE
```

### DIFF
--- a/templates/rhel9.tpl.yaml
+++ b/templates/rhel9.tpl.yaml
@@ -3,7 +3,7 @@ kind: Template
 metadata:
   name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
   annotations:
-    openshift.io/display-name: "Red Hat Enterprise Linux 9.0 VM"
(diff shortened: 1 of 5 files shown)


templates/rhel9.tpl.yaml
```diff
@@ -3,7 +3,7 @@ kind: Template
 metadata:
   name: {{ os }}-{{ item.workload }}-{{ item.flavor }}
   annotations:
-    openshift.io/display-name: "Red Hat Enterprise Linux 9.0 VM"
+    openshift.io/display-name: "Red Hat Enterprise Linux 9 VM"
```
dominikholler, line 6: I like this idea, but for consistency reasons this should be changed also for the RHEL 6, 7 and 8 template.

dominikholler (changes requested):
Thanks for the commit!
Would you please include a reference to the bug https://bugzilla.redhat.com/2143558 in the commit message? This would help us to get the change back into 4.11.

PR #504, opened 2023-01-03
decrease dv size of windows in e2e tests

ksimon1:
decrease dv size of windows in e2e tests
This change should reduce time needed to copy original windows image

**Release note**:
```
NONE
```


ksimon1:
/close
I will do it differently (reduce image size)

PR #505, opened 2023-01-10
switch to ssd storageClass in windows tests

ksimon1:
switch to ssd storageClass in windows tests

the non ssd storageclass is causing timeout on GCP OCP 4.12 windows tests. This commit changes the default storageClass to ssd.

**Release note**:
```
NONE
```


0xFelix (approved):
/approve
/lgtm

PR #506, opened 2023-01-11
fix: improve the condition for switching storageclass in e2e test

ksimon1:
fix: improve the condition for switching storageclass in e2e test
fix: changed order of applied annotations for storageclass in e2e test

**Release note**:
```
NONE
```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -74,9 +74,9 @@ export CDI_VERSION=$(latest_version "containerized-data-importer")
 
 # switch to faster storage class for widows tests (slower storage class is causing timeouts due 
 # to not able to copy whole windows disk into cluster)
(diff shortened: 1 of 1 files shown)


0xFelix (approved):
/lgtm
/approve

ksimon1:
the win lanes are failing, because in ocp 4.11 there is one extra storageclass, which is not available in 4.12 and is marked as default. Due to the fact, that this commit updates only 2 storageclasses, the third one is still marked as default and after the run of this script there are 2 storageclasses, which are marked as default and thus datavolume is not imported, because it doesn't know which one to use. I am merging this PR, because it will help to unblock https://github.com/openshift/release/pull/35150 which will upgrade ocp to 4.12 and we will get rid of the problem with 2 default storageclasses

PR #510, opened 2023-02-06
Update osinfo-db to v20221130-45-g30013a03

ksimon1:
Update osinfo-db to v20221130-45-g30013a03
**Release note**:
```release-note
Update osinfo-db to v20221130-45-g30013a03
```


0xFelix (approved):
/lgtm

PR #520, opened 2023-03-20
fix: add bind immediate annotation to test datavolume

ksimon1:
fix: add bind immediate annotation to test datavolume

the default storageClass in e2e tests requests WFFC mode. Currently when original datavolume is created in e2e tests, it is in pending state, because there is no object which will consume it. This commit adds special annotation
cdi.kubevirt.io/storage.bind.immediate.requested
to windows and linux original datavolume,
which allows to import datavolume without consumer.

**Release note**:
```
NONE
```


fossedihelm (approved):
Great! Thanks

PR #542, opened 2023-08-01
fix: custom configuration of crio in testing cluster

ksimon1:
our e2e are not working due to permission issue in crio.
https://github.com/cri-o/cri-o/pull/7192 should revert the change.
Until that revert PR is released and available in ci envs, we have to
set special configuration on crio on all nodes
**Release note**:
```
NONE

```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -72,15 +72,13 @@ export KUBEVIRT_VERSION=$(latest_version "kubevirt")
 # Latest released CDI version
 export CDI_VERSION=$(latest_version "containerized-data-importer")
 
-# switch to faster storage class for widows tests (slower storage class is causing timeouts due 
-# to not able to copy whole windows disk into cluster)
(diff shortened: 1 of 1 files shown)


automation/test.sh
```diff
@@ -70,11 +70,12 @@ function latest_version() {
 export KUBEVIRT_VERSION=$(latest_version "kubevirt")
 
 # Latest released CDI version
-export CDI_VERSION=$(latest_version "containerized-data-importer")
+#export CDI_VERSION=$(latest_version "containerized-data-importer")
```
jcanocan, line 73: If it's no longer required, we should drop this line to keep the file clean.

ksimon1: This is just testing PR, which should investigate where is the problem in CI. This change will not be part of result PR

ksimon1:
This PR is no longer needed

PR #546, opened 2023-08-23
feat: add podman to builder image

ksimon1:
feat: add podman to builder image

To make ci work again, we will need to have podman in test env. Podman will help to download VM image and push it to internal ocp registry

**Release note**:
```
NONE

```


ksimon1:
Mergin without CI, since it is not working. This PR should help in this PR https://github.com/kubevirt/common-templates/pull/542

PR #568, opened 2024-02-26
Update osinfo-db to v20231215-31-g87c980b1

ksimon1:
Update osinfo-db to v20231215-31-g87c980b1
**Release note**:
```release-note
Update osinfo-db to v20231215-31-g87c980b1
```


0xFelix (approved):
/approve

PR #583, opened 2024-06-03
feat: use spec.runStrategy instead of spec.running

### LINKED ISSUE #581: Replace spec.running with spec.runStrategy in all templates
We have templates configured with `spec.running` instead of `spec.runStrategy`. @iholder101 mentioned that `spec.running` will be deprecated and it should be replaced by `spec.runStrategy`.

Also, from the KubeVirt documentation [1] it's better to have `spec.runStrategy` because of:
```
There are cases when a user would like the ability to shut down a guest from inside the virtual machine.
With spec.running: true, KubeVirt would immediately restart the VirtualMachineInstance.
```
[1] https://kubevirt.io/user-guide/compute/run_strategies/#runstrategy

codingben:

VM templates are using spec.running and should use spec.runStrategy instead because spec.running is
planned to be deprecated soon.

Fixes #581

**Release note**:

```release-note
Use spec.runStrategy instead of spec.running
```

### DIFF
--- a/templates/centos-stream8.tpl.yaml
+++ b/templates/centos-stream8.tpl.yaml
@@ -85,7 +85,7 @@ objects:
           kind: DataSource
           name: ${DATA_SOURCE_NAME}
           namespace: ${DATA_SOURCE_NAMESPACE}
-    running: false
+    runStrategy: Halted
     template:
       metadata:
         annotations:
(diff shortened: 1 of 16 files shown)


0xFelix (commented):
/approve

PR #594, opened 2024-07-26
fix: make unit-tests can be called more than once

nestoracunablanco:
To enable running the unit tests multiple times, the existing templates are deleted before creating new ones. This approach allows for the successful execution of unit tests even if the templates already exist in the cluster.

If the templates do not exist, the 'oc delete' command would fail, which is why '|| true' is used.

**What this PR does / why we need it**: ability to run unit tests consecutively

Fixes #

**Release note**:

```release-note NONE

```

### DIFF
--- a/automation/unit-tests.sh
+++ b/automation/unit-tests.sh
@@ -5,6 +5,8 @@ set -ex
 templates=$(ls dist/templates/*)
 namespace="kubevirt"
 
+oc delete namespace ${namespace} || true
+
 echo "Testing fresh installation..."
 oc apply -f - <<EOF
 apiVersion: v1


jcanocan:
Nice catch @nestoracunablanco! These small details are always tricky.
/lgtm

automation/unit-tests.sh
```diff
@@ -25,6 +25,8 @@ for template in $templates; do
   oc process -f "$template" NAME=test DATA_SOURCE_NAME=test || exit 1
 done
 
+echo "generate all templates after cleaning up first"
+oc delete -f dist/templates || true
```
0xFelix, line 29: This would not delete templates which differ from the ones in `dist/templates` (see L51). What about deleting the namespace that was created above after a test run?

nestoracunablanco: Done.

0xFelix (commented):
/approve

PR #595, opened 2024-07-29
fix: verification of the python openshift requirement

nestoracunablanco:
The unit-tests.sh script may fail at an advanced stage if the python openShift module is not installed. Failing at the initial step alerts the user to this requirement.

**What this PR does / why we need it**: Fail early

Fixes #

**Release note**:

```release-note NONE

```

### DIFF
--- a/automation/unit-tests.sh
+++ b/automation/unit-tests.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -ex
 
+# dependency check
+pip show -q openshift
+
 #syntax check
 templates=$(ls dist/templates/*)
 namespace="kubevirt"
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -2,7 +2,7 @@
 FROM quay.io/fedora/fedora:latest
 
 # Install dependencies and tools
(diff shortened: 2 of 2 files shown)


automation/unit-tests.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -ex
 
+# dependency check
+pip show openshift
```
akrejcir, line 5: Can you make it quiet, so if openshift is installed, it would not print package information into the log?
```bash
pip show -q openshift
```

nestoracunablanco: Done

0xFelix (commented):
/approve
/ok-to-test

nestoracunablanco:
The change failed because the pip command could not be found in the container. After reviewing the following [file](https://github.com/openshift/release/blob/master/ci-operator/config/kubevirt/common-templates/kubevirt-common-templates-master.yaml#L8), I believed my last commit would resolve the issue. Unfortunately, that was not the case; it’s possible that the Dockerfile is being sourced from the main branch.
My plan is to close this pull request, create a new one with the necessary changes to the Dockerfile, and verify that it works correctly afterward.

PR #604, opened 2024-08-19
fix: e2e-tests preconditions in makefile

nestoracunablanco:
The preconditions for the end-to-end (e2e) tests were previously embedded within the script. This update ensures that unnecessary oc commands are not executed in the event of an error during the generation process.

**What this PR does / why we need it**: fail faster

Fixes #

**Release note**:

```release-note NONE

```

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -127,10 +127,6 @@ then
     export CLUSTERENV=$ocenv
 fi
 
-git submodule update --init
-
-make generate
-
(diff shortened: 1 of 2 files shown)


nestoracunablanco:
@jcanocan @0xFelix I believe that to enhance the reliability of this script, we should consider adding some wait conditions. However, for now, it seems sufficient for my needs. I will simply drop the PR and revisit later if necessary

PR #610, opened 2024-09-09
fix: remove tiny size from e2e test for fedora

ksimon1:
fix: remove tiny size from e2e test for fedora

the fedora tiny size was removed and e2e test is failing because it is trying to create vm with it.

**Release note**:
```
NONE

```


0xFelix (approved):
/lgtm

ksimon1:
Merging, the failing tests are not caused by this change

PR #612, opened 2024-09-12
fix: use spec.storage in DV in e2e tests

ksimon1:
fix: use spec.storage in DV in e2e tests

Switch to spec.storage in e2e tests DVs.
This should fix the 500 error we are experiencing during the
clone of the VM disk.

**Release note**:
```
NONE
```


jean-edouard (approved):
Thank you!

0xFelix (approved):
/approve
/lgtm

PR #616, opened 2024-09-19
feat: disable Centos 6 on s390x architecture

nestoracunablanco:
This commit disables the generation of templates for the s390x architecture, as there are no references to CentOS 6 in the osinfo-db for this architecture.

Fixes #

**Release note**:

```release-note NONE

```

### DIFF
--- a/automation/test-linux.sh
+++ b/automation/test-linux.sh
@@ -7,6 +7,11 @@ namespace="kubevirt"
 ocenv="OC"
 k8senv="K8s"
 
+if [ "$TARGET" == "centos6" ] && [ "$(uname -m)" == "s390x" ]; then
+  echo "CentOS 6 does not support s390x."
+  exit 1
+fi
+
 image_url=""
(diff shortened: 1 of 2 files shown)


automation/test-linux.sh
```diff
@@ -7,6 +7,11 @@ namespace="kubevirt"
 ocenv="OC"
 k8senv="K8s"
 
+if [ "$TARGET" == "centos6" ] && [ "$(uname -m)" == "s390x" ]; then
+  echo "CentOS 6 does not support s390x."
+  exit 1
```
ksimon1, line 12: This will always fail the e2e tests if run on s390X.
Wouldn't be better to just print out warning and exit with 0?

nestoracunablanco: It is intentional for the CentOS 6 test to fail when run on s390x, as this Linux distribution, according to my investigation in osinfo-db, does not provide images for this architecture.

PR #618, opened 2024-09-20
fix: fix minimum requirements in win2k25 template

ksimon1:
fix: fix minimum requirements in win2k25 template

The win2k25 requirements https://learn.microsoft.com/en-us/windows-server/get-started/hardware-requirements?tabs=cpu#components

**Release note**:
```
NONE

```


0xFelix (approved):
/approve
/lgtm

PR #619, opened 2024-09-23
[release-v0.31] fix: revert #583

ksimon1:
fix: revert #583
QE team is not prepared for this change and tests are failing due to that.

**Release note**:
```
 fix: revert #583
```


0xFelix (approved):
/approved

PR #621, opened 2024-10-14
feat!: remove centos7 and centos 8 stream templates

ksimon1:
feat!: remove centos7 and centos 8 stream templates

These OSes are no longer supported

Fixes 
https://issues.redhat.com/browse/CNV-48104

**Release note**:
```
feat!: remove centos7 and centos 8 stream templates

```


nestoracunablanco:
@ksimon1 first of all, I think it's great that you're removing some distributions, especially those that don't support many architectures. However, I noticed that CentOS 6 is not being deleted. Is there a specific reason for this?

ksimon1:
@nestoracunablanco yes, there is still business need for keeping the centos6 template in place.

automation/test-linux.sh
```diff
@@ -80,7 +80,7 @@ if [[ $TARGET =~ opensuse.* ]]; then
   workloads=("server")
 fi
```
geetikakay, line 82: since we are touching this file , should we cleanup centos6.*

generate-templates.yaml
```diff
@@ -171,35 +142,6 @@
...
-      majorrelease: centos7
-      oslabels:
-       - centos7.0
-      osinfoname: "{{ oslabels[0] }}"
-      cloudusername: centos
-      containerdisk_urls: "{{ centos7_containerdisk_urls }}"
-      image_urls: "{{ centos7_image_urls }}"
-
   - name: Load CentOS 6 versions
     set_fact:
```
geetikakay, line 146: nit: maybe we can remove centos6?

PR #573, opened 2024-04-10
Set architecture to amd64 across all templates

lyarwood:
/cc @0xFelix 
/cc @ksimon1 

**What this PR does / why we need it**:

The current set of templates are only verified on amd64 so to avoid users attempting to run them on aarch64 etc set the architecture within the VirtualMachineInstanceSpec.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
The architecture of each `VirtualMachine` is now  set to `amd64` within the `VirtualMachineInstanceSpec`.
```

PR #624, opened 2024-10-17
feat: multiarch fedora template.

nestoracunablanco:
Fedora offers images for a variety of architectures, which can be included as a template parameter. Within the template, there is a specific parameter for defining the architecture.

For the sake of simplicity, I have currently set the host architecture.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/generate-templates.yaml
+++ b/generate-templates.yaml
@@ -229,8 +229,8 @@
 
   - name: Load Fedora containerdisk and image urls
     set_fact:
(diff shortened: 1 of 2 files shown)


0xFelix:
@ksimon1 Can you remind me why we need this field at all in the template?

@nestoracunablanco Did you check all fields enabled in the template work with s390x?

nestoracunablanco:
@0xFelix, next Monday, I will likely have access to a machine capable of running the Fedora end-to-end tests on s390x. So far, I have verified that the generated template includes the link to the s390x images and correctly points to the s390x architecture. Additionally, I tested to ensure that it does not break the amd64 configuration, which you will probably notice in the CI.

ksimon1:
> @ksimon1 Can you remind me why we need this field at all in the template?
> 
https://github.com/kubevirt/common-templates/pull/573
As @lyarwood described in PR, the templates are only tested against amd64 machines

PR #636, opened 2024-11-15
fix: increase timeout for end-to-end tests

nestoracunablanco:
In some cases, the current timeout of 600 seconds is insufficient for starting a virtual machine (VM). To address this issue, we propose increasing the timeout to 900 seconds, which should provide adequate time for the VM to start successfully.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/automation/test-linux.sh
+++ b/automation/test-linux.sh
@@ -44,7 +44,7 @@ spec:
         storage: 30Gi
 EOF
 
-timeout=600
+timeout=900
 sample=10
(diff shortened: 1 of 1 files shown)


jcanocan:
As a follow-up, we should understand why VMs are taking so long, 600 seconds should be enough to start a VM.
/lgtm

0xFelix (approved):
/approve

PR #649, opened 2024-11-27
feat: add suffix to the s390x templates

nestoracunablanco:
This change also involves renaming the templates themselves. The end-to-end (e2e) tests require minimal adjustments.

Fixes #

**Release note**:

```release-note
NONE
```


generate-templates.yaml
```diff
@@ -62,7 +62,7 @@
   - name: Generate RHEL 8 templates
     template:
       src: rhel8.tpl.yaml
-      dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}.yaml"
+      dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}{{ '' if target_arch == 'x86_64' else '-' + target_arch }}.yaml"
```
0xFelix, line 65: Do we want to separate the generated templates also by path?

nestoracunablanco: We can do it if you’d like, but this would require changes to the following commits as well. Ultimately, the release will consist of three files: one containing all architectures and one for each supported architecture.

What are your thoughts, @akrejcir, @jcanocan, @ksimon1, @lyarwood?

0xFelix: > Ultimately, the release will consist of three files: one containing all architectures and one for each supported architecture.

As long as it integrates nicely with SSP we can keep it that way.

ksimon1:
Would you please specify in commit message which templates have the s390x version?
Other s390x templates (like windows) will be introduced in next PRs?
/cc @fossedihelm

fossedihelm:
/hold
I think rhel9.tpl.yaml is missing a change at line 54
Also why don't we just pass the suffix as var from `generate-templates.yaml`:
```
    vars:
      [other vars...]
      suffix: "{{ '' if target_arch == 'x86_64' else '-' + target_arch }}"
```
And then just do"
```
  name: {{ os }}-{{ item.workload }}-{{ item.flavor }}{{ suffix }}
```
?

nestoracunablanco:
@ksimon1 comment updated.
@fossedihelm nice catch. Will do.

generate-templates.yaml
```diff
@@ -27,7 +29,7 @@
   - name: Generate RHEL 9 templates
     template:
       src: rhel9.tpl.yaml
-      dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}.yaml"
+      dest: "{{ playbook_dir }}/dist/templates/{{ os }}-{{ item.workload }}-{{ item.flavor }}{{ '' if target_arch == 'x86_64' else '-' + target_arch }}.yaml"
```
fossedihelm, line 32: You can use the suffix here too :)

fossedihelm (approved):
Thank you!

akrejcir:
Can you consider adding a new label on the template that would contain the architecture? Then we can easily filter templates based on it.

templates/centos-stream9.tpl.yaml
```diff
@@ -49,13 +49,14 @@ metadata:
 {% if item.default %}
     template.kubevirt.io/default-os-variant: "true"
 {% endif %}
+    architecture: "{{ 'amd64' if target_arch == 'x86_64' else target_arch }}"
```
0xFelix, line 52: The label should be prefixed with `template.kubevirt.io/`

nestoracunablanco: Done.

0xFelix (commented):
/approve

fossedihelm (approved):
/lgtm
/hold
@akrejcir want to take a final look?

PR #664, opened 2025-01-27
Update osinfo-db to v20250124

ksimon1:
Update osinfo-db to v20250124
**Release note**:
```release-note
Update osinfo-db to v20250124
```


0xFelix (approved):
/approve

PR #669, opened 2025-03-11
Enable arm64 builds

fossedihelm:
Enables arm64 support for common templates

Fixes #

**Release note**:

```release-note
arm64 support
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,6 +38,7 @@ jobs:
             ./dist/common-templates-${{ steps.get_release.outputs.tag_name }}.yaml
             ./dist/common-templates-amd64-${{ steps.get_release.outputs.tag_name }}.yaml
             ./dist/common-templates-s390x-${{ steps.get_release.outputs.tag_name }}.yaml
+            ./dist/common-templates-arm64-${{ steps.get_release.outputs.tag_name }}.yaml
 
       - name: Install golang
         uses: actions/setup-go@v2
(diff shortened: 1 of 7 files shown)


fossedihelm:
/hold

@0xFelix I am not sure which templates I should enable.

0xFelix:
@fossedihelm We have common instancetypes for fedora/rhel9/rhel10 on ARM. So I suggest that as a start. Maybe add CentOS Stream 9 and 10 and Ubuntu too?

fossedihelm:
Thank you @0xFelix.
So, currently I enabled arm64 for:
- rhel8
- rhel9
- centOS Stream 9
- fedora
- ubuntu

rhel10 and centOS Stream 10 are not included in the templates (maybe we should add them in a separate PR).
Thanks

0xFelix (commented):
Looking good so far, but IIRC SecureBoot and possibly even EFI are not supported aarch64, similar to s390x. I recommend to verify that first. You can find conditionals for s390x in the templates already.

templates/rhel8.tpl.yaml
```diff
@@ -45,7 +45,7 @@ metadata:
 {% if item.default %}
     template.kubevirt.io/default-os-variant: "true"
 {% endif %}
-    template.kubevirt.io/architecture: "{{ 'amd64' if target_arch == 'x86_64' else target_arch }}"
+    template.kubevirt.io/architecture: "{{ 'amd64' if target_arch == 'x86_64' else 'arm64' if target_arch == 'aarch64' else target_arch }}"
```
0xFelix, line 48: There are no common-instancetypes for rhel8 on ARM. I'd keep that out for now.

0xFelix: You can drop changes to `rhel8.tpl.yaml`

fossedihelm: Right! Thank you!

0xFelix (commented):
/approve

Thanks

0xFelix:
We could override Windows lanes as they are not touched by this PR anyway until https://github.com/kubevirt/common-templates/pull/672 is fixed.

0xFelix:
The tests in this repo need to be adapted to check if the guest agent is connected instead of using port forwarding, which is broken on the latest OCP atm.

stu-gott:
@fossedihelm @0xFelix Windows templates aren't currently mandated or required.

0xFelix:
@stu-gott The Windows lanes still block this PR even though they are not required for ARM. CI/CD in this repository could use a cleanup / overhaul in general.

PR #672, opened 2025-03-14
fix: Windows e2e tests

ksimon1:
fix: Windows e2e tests
Windows images now uses only 4.99 tag

fix: use AgentConnected status condition to recognize if Windows VM is up
removed retry of mechanism of starting the VM, since it is not require
any more.
Changed the way how to recognize if Windows vms are up are ready. Tests
now uses AgentConnected status condition which is set, when vm is up.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/automation/test-windows.sh
+++ b/automation/test-windows.sh
@@ -26,7 +26,7 @@ elif [[ $TARGET =~ windows2025.* ]]; then
 fi
 
 source_name="${TARGET}-original"
-version=$(oc version -o json | jq -r '.openshiftVersion | split("\\."; null)[:2]|join(".")')
+version="4.99"
 
 oc apply -n ${namespace} -f - <<EOF
 apiVersion: image.openshift.io/v1
--- a/automation/test-windows.sh
+++ b/automation/test-windows.sh
@@ -4,7 +4,6 @@ set -ex
 
 namespace="kubevirt"
 template_name="windows2k22"
-username="Administrator"
 
 sizes=("medium" "large")
 workloads=("server" "highperformance")
@@ -67,10 +66,8 @@ spec:
   managedDataSource: ${source_name}
 EOF
 
-timeout=2000
 hour_timeout=3600
-sample=10
-current_time=0
+fifteen_minutes_timeout=900
 
 oc wait --for=condition=UpToDate --timeout="${hour_timeout}s" "dataImportCron/${source_name}" -n "${namespace}"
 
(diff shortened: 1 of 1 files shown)


ksimon1:
Failing win10/11 are caused by not able to ssh to VMs. The Windows vms are up and sshd service is running (checked), but they are unaccessible from outside. Not sure if it is caused by Kubevirt or some other issue.

ksimon1:
@fossedihelm, @0xFelix, @jcanocan can you please review?

automation/test-windows.sh
```diff
@@ -88,51 +84,18 @@ run_vm(){
...
-    #no error were observed, the vm is running
-    if ! $error ; then
-      running=true
-      break
-    fi
-  done
-
-  set -e
+  oc wait --for=condition=Ready --timeout="${hour_timeout}s" "vm/${vm_name}" -n "${namespace}"
+  oc wait --for=condition=AgentConnected --timeout="${hour_timeout}s" "vm/${vm_name}" -n "${namespace}"
```
0xFelix, line 96: Do we really want this high timeout here?

automation/test-windows.sh
```diff
@@ -88,51 +84,18 @@ run_vm(){
...
 
-  if ! $running ; then
-    exit 1 
-  fi
+  delete_vm $vm_name $template_name
```
0xFelix, line 98: VM is not deleted in case of failure?

0xFelix (commented):
Can you please update the PR description / title?

automation/test-windows.sh
0xFelix: Right, right now this is not using a single cluster to run multiple tests. Might be an improvement for the future though? Spinning up many clusters is costly and fails often times due to exhausted resources.

automation/test-windows.sh
```diff
@@ -88,51 +85,18 @@ run_vm(){
...
-    delete_vm $vm_name $template_name
-    #no error were observed, the vm is running
-    if ! $error ; then
-      running=true
-      break
-    fi
-  done
-
-  set -e
+  oc wait --for=condition=Ready --timeout="${fifteen_minutes_timeout}s" "vm/${vm_name}" -n "${namespace}"
```
0xFelix, line 96: IMO we can keep the hour timeout for the Ready condition, but once the VM runs it should boot within the shorter timeout.

ksimon1: the vm should be ready in just few minutes, so 15 minutes is ok

0xFelix: We had a timeout of 1 hour before?

ksimon1: yes, because we didn't use ssd as storage. But since then, the import, start, test should be done in less than 15 minutes each

0xFelix (commented):
/approve

PR #687, opened 2025-06-02
Update osinfo-db to v20250124-37-g2a374acd

ksimon1:
Update osinfo-db to v20250124-37-g2a374acd
**Release note**:
```release-note
Update osinfo-db to v20250124-37-g2a374acd
```


0xFelix (commented):
/approve

PR #701, opened 2025-12-01
Update osinfo-db to v20250606-38-gf008aba7

ksimon1:
Update osinfo-db to v20250606-38-gf008aba7
**Release note**:
```release-note
Update osinfo-db to v20250606-38-gf008aba7
```


0xFelix (commented):
/approve

